PR #52, opened 2016-05-18
Add left right rotation style

stevekrouse:


stevekrouse:
Looks great, Sally! Can you update the commit hash in the README script tag and the commit hash in the jsbin demo (this will mean cloning the jsbin demo and pasting in a new link)?

PR #60, opened 2016-05-19
repeat instant

stevekrouse:


stevekrouse:
Great!

@sallymatson, can you explain what `repeatInstant` is and how it's bad for animation?

Don't worry about the hash

stevekrouse:
@sallymatson your code says it's `Woof.repeatInstant` but the REAME says it's `project.repeatInstant`

PR #160, opened 2016-07-14
Fix image links in README section headings

adidahiya:


stevekrouse:
Hey @adidahiya! Thanks so much for this!

I think my Github email notification settings are messed up because I didn't notice this till just now (after I already fixed this bug)

Thanks again! I appreciate that you're looking out for me!

PR #275, opened 2016-12-15
Backdrop div

jazzyjackson:
I did this!

Instead of renamed backdropCanvas and backdropContext all over the place, they all refer to the same div node with id backdrop.


jazzyjackson:
Made changes from my fork, cancelling this request

PR #276, opened 2016-12-16
Backdrop div

jazzyjackson:
Before these changes, the background image or color was being painted as a bitmap to a backdrop canvas.

Instead, for ease of applying CSS rules for tiling the image, that canvas is simply a DIV which takes its style from thisContext.height and thisContext.width, just like the canvases.

methods that are added alongside setBackdropUrl and setBackdropColor include setBackdropStyle and setBackdropRepeat

usage of setBackdropStyle would include
setBackdropStyle('cover')
setBackdropStyle('50%')
setBackdropStyle('30px 50px')

So long as the arguments follow this style, they will be plugged right into the background-size property of the backdrop div. If not, an informative error will be thrown.

This doesn't support the advanced usage for multiple background images. Only takes 1 or 2 space separated arguments, no commas.

setBackdropRepeat also has a whitelist of valid CSS property names, which if matches, background-repeat property of the backdrop div will be set.


stevekrouse:
Wowza 💥, this is so thorough!! Really appreciate that you added the hints and also renamed the variables -- going the extra mile for sure! Thanks @jazzyjackson! Huge help 👍 

Would you like to take a stab at [adding the new commands you created to the interactive documentation](https://github.com/stevekrouse/WoofJS/blob/master/docs/docs/index.html#L316)? 

(We didn't really go over this code, so I'd be happy to do this part if you'd rather. It's built in VueJS which is similar to React so you could probably figure it out. Let me know if you have any questions about it.)

jazzyjackson:
Np! I'll check out Vue js. If you tell me where the result of those objects ends up I'm sure I can figure it out.

In the meantime, I've found a hack to get codemirror's gutter to extend the whole height, depends on how adverse you are to the !important selector.

docs > codemirror > codemirror.css

.CodeMirror-gutters {
    min-height: 100% !important;
...

Whenever a new row is added to the codemirror, the height is set as an attribute on the element. I'm going to follow a debugger to see if I can find out where... but the above style overrides it and it looks nice so ¯\_(ツ)_/¯

jazzyjackson:
Oh! I see where the interactive docs end up

Since it only makes sense to make those changes along with this pull request, should I make changes on this branch and make another pull request? or just make changes on master and put a pull request on that?

stevekrouse:
Great, good work figuring out that code!

Agreed, I think all these changes (minus the gutter changes) belong in the same pull request. As far as I know, if you add more commits to your backdropDiv branch and push them, they'll appear as additional commits on this pull request (without needing to create a new pull request). Do you want to try that?

jazzyjackson:
Oh, so it does, nifty! There are your example objects :)

stevekrouse:
Amazing! Thanks again! 

I want to test it out with a bunch of different student code and variations before merging in such a large change, so I hope to get to it in the next few days. Bear with me 😁

stevekrouse:
This pull request is really close! I don't know an easy way to pull the changes from yourjazzyjackson/WoofJS branch so the easiest thing I could think of was to merge this pull request into a non-master branch in stevekrouse/WoofJS so I could easily pull it and make changes.

Hope that's alright. If you know a better way for me to do this, I'd love to hear about it :)

stevekrouse:
Again, thanks for the amazing work, Colten! Hope you don't mind me taking it the final few inches it needed to get merged in.

If you were curious, I updated the default to 100% 100% no-repeat so that existing WoofJS projects would be unaffected by this change, made a few minor docs adjustments, and updated the woof.js hash that raw git pulls. You can see the changes in these commits: 1e9b9a5e25ec87307efbe17924a09c4e0865fe6e, 7f054113e88c59c36cf85246a5bdfa46ba45a098, e37d91636b0d22ea890652a2244bbc610d2f3f19, ddefe0282d323aa3e49dbc96a8e1cbe76601009f

jazzyjackson:
Excellent, glad it could be integrated.

I want to ask, what do the part time teaching positions look like?

I'm having sudden second thoughts about committing myself full time to
someone else's projects, so I might be looking for a way to subsidize
myself working on my own ideas.

- Colten

PR #289, opened 2017-02-10
remove all imgur and postimage files links

madduccino:
Aside from links, references to Imgur and Postimage were found here and left as is:

/WoofJS/docs/index.html:
402: description: "Use postimage.org to upload your own image",

/WoofJS/dist/woof.js:
1015: console.warn("You have an image at an untrusted URL. Consider uploading to Imgur and using https.");

/WoofJS/src/woof.es6.js:
685: console.warn("You have an image at an untrusted URL. Consider uploading to Imgur and using https.")


stevekrouse:
Great work, @madduccino!

PR #290, opened 2017-02-10
home page redesign: no toggle buttons, x buttons, logo and color change

nicolekelner:
No more tooltip


create.html
```diff
@@ -445,7 +496,7 @@
     }
     
     // set the text in the editor
-    var editorVal = "setBackdropURL('https://i.imgur.com/q6Bqraw.jpg')\n\n\n"
+    var editorVal = "setBackdropURL('http://i.imgur.com/kfyszaH.png')\n\n\n"
```
stevekrouse: As long as you're changing this, can  you change it to a non imgur link?

create.html
```diff
@@ -226,50 +229,78 @@
...
-            <iframe src="./docs/index.html" id="scratch" style="height:100%; width:100%;"></iframe> 
-        </div>
-        
-        <div v-show="projectsPane" id="projects" style="width:25%; height:100%; flex: 1.2 1 25%; word-wrap: break-word; ">
-          <iframe id="projectsIframe" src="http://coding.space/woof/index.html" style="height:100%; width:100%"></iframe> 
+            
+            
+        <div  v-show="docsbar"  @click="docsbar= !docsbar"  :style="{display: Minimize ? 'inline-block, display' : 'none, flex' } " style=" width:20px; height:100%; display:flex; flex-direction:column; justify-content: center;  background-color:#7F68D5" >
+              <img src="./images/documentation.png" style=" width: 15px; height:125px;" >
+            </div>
```
stevekrouse: the formatting of this code is off

create.html
```diff
@@ -226,50 +229,78 @@
...
+       
       </span>
     </div>
  
     <div id="container" style="diplay:flex; ">
-        <div v-show="outputPane" id="output" style="height:100%; resize: none; flex: 1.2 1 25%;">
+       
+       <div  v-show="previewbar"  @click="previewbar= !previewbar"  :style="{display: Minimize ? 'inline-block, display' : 'none, flex'}" :style="{display: Minimize ? 'flex-direction' : 'column'}" style=" width:20px; height:100%; display:flex; flex-direction:column; justify-content: center;  background-color:#4875D1" >
+              <img src="./images/preview.png" style=" width: 15px; height:65px;" >
+            </div>
```
stevekrouse: the formatting of this code is off

create.html
```diff
@@ -226,50 +229,78 @@
...
+              <img src="./images/preview.png" style=" width: 15px; height:65px;" >
+            </div> 
+            <div v-show="outputPane" :style="{display: !previewbar ? 'inline-block' : 'none'}"  id="output" style="height:100%; resize: none; flex: 1.2 1 25%;">
+              <div id ="PreviewNav" style="text-align:center; background-color:#4875D1; color:white; font-size:1.1em style; display:flex; justify-content: space-between;"> 
+                 <span style="margin-left:8px;  margin-top:4px">Preview</span>
+                 <span style="margin-right:8px" >
+                     <img onclick="popout()" id="pop" style="margin: 5px; width: 18px; height:18px" src="./images/pop.png">
+                     <img @click="previewbar = !previewbar"  style="margin: 5px; width: 16px; height:18px" src="./images/minimize.png">
+                 </span>
+              </div>
```
stevekrouse: the formatting of this code is off

create.html
```diff
@@ -226,50 +229,78 @@
...
+        <div v-show="projectbar"  @click="projectbar= !projectbar"  :style="{display: Minimize ? 'inline-block, display, margin-left' : 'none, flex,.5px' } " style="width:20px; height:100%; display:flex; flex-direction:column; justify-content: center;  background-color:#AD5FC2"   @click="projectsPane = !projectsPane" :class="{'ios-container-disabled': !projectsPane}">
+              <img src="./images/tutorials.png" style=" width: 15px; height:78px; margin: auto" >
+            </div> 
+            <div v-show="projectsPane" :style="{display: !projectbar ? 'inline-block' : 'none'}"  id="projects" style="width:25%; height:100%; flex: 1.2 1 25%; word-wrap: break-word; ">
+               <div id ="ProjectsNav" style="text-align:center; background-color:#AD5FC2; color:white; font-size:1.1em; display:flex; justify-content: space-between; border:none"> 
+                    <span style="margin-left:8px; margin-top:4px">Tutorials</span>  
+                    <span style= "margin-right:8px">
+                        <img @click="projectbar = !projectbar"  style="margin: 5px; width: 16px; height:18px" src="./images/minimize.png">
+                    </span>
+              </div>
```
stevekrouse: the formatting of this code is off

create.html
```diff
@@ -324,13 +358,26 @@
           // whether or not the code can be "cleaned" or if its already pristine
           return !this.error && this.code != js_beautify(this.code, {space_after_anon_function: true, indent_size: 2, indent_with_tabs: false })
         },
+        
+        Minimize: function() {
```
stevekrouse: Unless I am very confused, I don't think this function does anything productive.

You reference it in your code, like `:style="{display: Minimize ? 'inline-block, display' : 'none, flex' } "`. Given that you don't call the Minimize function, which would require parenthesis `Minimize()`, all you're doing is checking if the Minimize function exists, which it always does. 

This is equivalent to writing `:style="{display: true ? 'inline-block, display' : 'none, flex' } "`.

Which is equivalent to writing `style="display: inline-block"` with no vue properties. Just regular CSS.

I'm also not sure why you have `, display` and `, flex` in your expression. That doesn't do anything either.

Please delete this function and remove all references to Minimize from your code :)

stevekrouse:
This is really great, @nicolekelner! I'm very excited about the new look.

Besides the code-specific comments I made, I have a few design ideas.

1. The vertical menus are a little thin. Obviously we don't want them to take up too much space but just a few pixels wider could do the trick. 


2. What do you think of doing away with the X buttons and having it be that when you click on the top bar of each section, it opens and closes that section? It's definitely less clear but I keep finding myself using the bars like that and am bummed when it doesn't work like that. 


3. This can be a separate pull request / issue but it could be fun to animate the opening and closing of the section menus. Basically, you could make the section heading and the vertical divider be the same element that you rotate down. You can play with the general idea on [codepen](http://codepen.io/pen/)
by dragging their section menus closed. Don't let this derail you but I thought you could have fun monkeying with it. Instead of using vertical pictures of the text, you would use css rotation somehow and then add a transition effect to it.

PR #298, opened 2017-02-21
make docs page responsive

madduccino:


docs/docs.css
```diff
@@ -140,8 +155,13 @@ pre {
     cursor: help;
 }
 
-#searchbar {
-    width: 80%;
+.search {
```
stevekrouse: Seems like you can remove this empty CSS selector

docs/index.html
```diff
@@ -13,41 +13,40 @@
...
+    <div id="body">
         <div :style="menuStyles" id="menus">
-
-            <div style="max-width: 100%" v-for="(blocks, menu) in searched(menus)">
-                <template v-if="menu != 'Sprites & Backgrounds'"> 
-                    <button v-show="open == undefined || open == menu || search" :class="{highlight: !search}" class="all" :id="menu.split(' ')[0]" @click="toggle(menu)">{{ menu }}</button>
-                    <div v-show="open == menu || search" class="list" v-for="block in blocks">
+            <div class="btn-group search">
+                <button v-show="open != undefined || search != ''" @click="search = ''; open = undefined" id="back" class="highlight btn btn-medium btn-info">Back</button>
+            <input  id="searchbar" type="search" class="form-control" placeholder="Search blocks..." v-model="search"></input>
```
stevekrouse: this line needs to be indented two spaces to the right

docs/docs.css
```diff
@@ -140,26 +140,70 @@ pre {
...
+    width: 100%;
     padding: 10px;
     border-radius: 5px;
     font-size: 16px;
-    margin: 10px;
     border: 1px solid black;
     outline: 0px;
 }
 
+#menus div:nth-child(3) {
```
stevekrouse: The way the docs page used to work is that the ordering of the menus was specified in the menus array  https://github.com/stevekrouse/WoofJS/blob/master/docs/index.html#L81

1) How did your code change that? Why did it mess up the natural ordering of this array?

2) Can you re-order the elements in the menus array in JavaScript instead of doing it here in CSS?

PR #311, opened 2017-02-28
background buttons 50%

nicolekelner:


stevekrouse:
Seems like this doesn't totally solve the problem when things are really squished...

screenshot 2017-03-01 at 10 51 39 am

Why don't you make them 100%?

screenshot 2017-03-01 at 10 53 21 am

PR #321, opened 2017-03-04
Sprite hint

nicolekelner:
Prompts user to change their variable if they leave default sprite1 and then changes in the code for all instances

### DIFF
--- a/create.html
+++ b/create.html
@@ -164,13 +164,6 @@
       margin-right: 5px;
     }
     
-     #suggestion {
-      display: none;
-      color: white;
-      margin-right: 5px;
-      background-color: #F2C94C;
-    }
-    
     .rotate {
       color: #FFFFFF;
       -webkit-transform: rotate(90deg);
@@ -335,20 +328,17 @@
         getID: getID,
         showError: function(){
           // If "sprite1" is in the editor, ask the user for how they'd like to rename sprite1
-           if (editor.getValue().includes('sprite1') && $('#logggedin').is(':visible')) {
-            var sprite_name = prompt("It looks like you have not renamed your your sprite! It is still set as the default 'Sprite1'.\n\nWhat is the name of your variable?", "unicornSprite");
-                if (sprite_name === null) {   
-                  return;}
-              
-            editor.setValue(editor.getValue().replace(/sprite1/g, sprite_name));
-            
-            
-            
+          if (this.error.message.includes('sprite1') && $('#logggedin').is(':visible')) {
+            var spriteName = prompt("It looks like you have not renamed your your sprite! It is still set as the default 'Sprite1'.\n\nWhat is the name of your variable?", "unicornSprite");    
+            if (spriteName !== null) {   
+                editor.setValue(editor.getValue().replace(/sprite1/g, spriteName));
             }
-            else 
+            
+            
+          }
+          else {
              alert(this.error.message)
-         
-          
+          }
         },
     
         
--- a/create.html
+++ b/create.html
@@ -606,26 +606,46 @@
        }		
     };
     
-    function tryRunningCode(){
+    function tryRunningCode() {
       app.code = editor.getValue()
       app.error = undefined
+    
       try {
-        var code = Babel.transform(editor.getValue(), { presets: ['es2015'] }).code
+        var code = Babel.transform(editor.getValue(), {
+          presets: ['es2015']
+        }).code
         code = code.replace("'use strict';", '')
+    
         try {
-          document.getElementById('preview').contentWindow.addEventListener("error", function(error) { app.error = error } );
-          
-          var script = document.createElement( "script" );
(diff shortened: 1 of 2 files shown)


stevekrouse:
View this pull request at:

https://cdn.rawgit.com/stevekrouse/WoofJS/d4ed0dec66f91e0d0f1a87333fe8808fbbb2ea6a/index.html

create.html
```diff
@@ -606,27 +606,47 @@
...
     };
     
-    function tryRunningCode(){
+    function tryRunningCode() {
       app.code = editor.getValue()
       app.error = undefined
+    
       try {
-        var code = Babel.transform(editor.getValue(), { presets: ['es2015'] }).code
+        var code = Babel.transform(editor.getValue(), {
```
stevekrouse: I looks like you formatted this block of code which makes it look like you edited a whole bunch of lines that you in fact didn't really change. Can you try to only do the sublime text format shortcut on only the lines of code that you actually are editing?

create.html
```diff
@@ -606,27 +606,47 @@
...
+            app.error = error
+          });
+          var script = document.createElement("script");
           script.type = "text/javascript";
           script.text = code
           document.getElementById('preview').contentDocument.body.appendChild(script)
-        } catch (e) {
-          app.error = e
+    
+          // If sprite1 is not changed, create popup for user to input variable name
```
stevekrouse: How about `// If "sprite1" is in the editor, ask the user for how they'd like to rename sprite1`

create.html
```diff
@@ -606,27 +606,47 @@
...
+          if (editor.getValue().indexOf('sprite1') != -1) {
```
stevekrouse: I prefer using the `includes()` function to `indexOf ~= -1` because it reads easier so it's more clear what you're doing.

https://www.w3schools.com/jsref/jsref_includes.asp

create.html
```diff
@@ -606,27 +606,47 @@
...
+            var sprite_name = prompt("Enter the name of your Sprite", "unicorn_sprite");
```
stevekrouse: Can you make the prompt message more clear? Maybe mention that we're changing `sprite1`...? 

Also, can you uncapitalize "Sprite"? 

Also, can you remove the underscore in "unicorn_sprite"? In JavaScript we do camel case, so it could be "unicornSprite".

src/woof.es6.js
```diff
@@ -1071,6 +1071,14 @@ Woof.prototype.Repeat = function(times, func, after) {
   };
 };
 
+//error if user does not change sprite1
+Woof.prototype.Sprite = function() {
```
stevekrouse: What does this code do?

create.html
```diff
@@ -160,9 +160,15 @@
     
     #error {
       display: none;
-      color: #a94442;
+      color: white;
       margin-right: 5px;
-      background-color: #f2dede;
+    }
+    
+     #suggestion {
```
stevekrouse: please remove this :)

create.html
stevekrouse: Again, given that you didn't actually edit any of this code, please put it back to how it was before. Let me know if you need help with that.

src/woof.es6.js
```diff
@@ -1070,6 +1070,8 @@ Woof.prototype.Repeat = function(times, func, after) {
     }
   };
 };
+
```
stevekrouse: Same goes for this code. Given that you didn't edit this file, there should be no newlines in the diff.

create.html
```diff
@@ -328,7 +334,21 @@
       methods: {
         getID: getID,
         showError: function(){
-          alert(this.error.message)
+          // If "sprite1" is in the editor, ask the user for how they'd like to rename sprite1
+           if (editor.getValue().includes('sprite1') && $('#logggedin').is(':visible')) {
+            var sprite_name = prompt("It looks like you have not renamed your your sprite! It is still set as the default 'Sprite1'.\n\nWhat is the name of your variable?", "unicornSprite");
```
stevekrouse: Please change `sprite_name` to `spriteName`

create.html
```diff
@@ -328,7 +334,21 @@
...
+                if (sprite_name === null) {
```
stevekrouse: This code is formatted improperly. Also, the logic could be more clear like this...

```javascript
if (spriteName !== null) {
  editor.setValue(...)
}

File: create.html [hunk 10]
```diff
@@ -328,7 +334,21 @@
       methods: {
         getID: getID,
         showError: function(){
-          alert(this.error.message)
+          // If "sprite1" is in the editor, ask the user for how they'd like to rename sprite1
+           if (editor.getValue().includes('sprite1') && $('#logggedin').is(':visible')) {
```
stevekrouse: Instead of checking if the editor contains `'sprite1'`, it probably makes sense to check if the `this.error.message.includes('sprite1 is not defined')`

Given that this Error is going to show up regardless of if the user is logged in, we don't need to check that anymore probably.

PR #322, opened 2017-03-04
adjust text editor and chat window settings

madduccino:
I tried a few things:

1) Added scroll past end plugin

2) Added more blank lines to text editor. I tried 60 lines at one point (is there a better way to do this than add 60 "/n"s?) to achieve the goal you were referring to (I believe) but the vertical scrollbar that appears as a result hijacks some of the text editor's limited space when all tabs are open. I still think additional lines might look good, though, and possibly provide users just an iota more direction, so I left it at 25.

3) Another issue the chat window creates besides leaving users wondering what to do when reaching the bottom is confusion over how to horizontally scroll when the line gets too long. I added line wrapping to eliminate this problem altogether.

4) I made the chat window slightly smaller and justified all the way right so it's less in your face but still visible.

5) Are the text editor gutters necessary? I commented them out because I feel like the more space on the editor, the better.

### DIFF
--- a/create.html
+++ b/create.html
@@ -61,6 +61,7 @@
   <link rel="stylesheet" href="./vendor/codemirror/lint.css">
   <script src="./vendor/codemirror/jshint.min.js"></script>
   <script src="./vendor/codemirror/lint.js"></script>
+  <script src="./vendor/codemirror/scrollpastend.js"></script>
   <script src="https://unpkg.com/vue@2.0.1/dist/vue.js"></script>
   <script data-presets="es2015" src="https://cdnjs.cloudflare.com/ajax/libs/babel-standalone/6.14.0/babel.min.js"></script>
   <script src="https://cdn.rawgit.com/beautify-web/js-beautify/1b52eb1f90daaefa6ff0fa7736f97fbfc58093c1/js/lib/beautify.js"></script>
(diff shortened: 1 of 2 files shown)


stevekrouse:
View this pull request at:

https://cdn.rawgit.com/stevekrouse/WoofJS/0dd5ddc9bac7623ed9d7258ba4ccb82872792786/index.html

stevekrouse:
Thanks for the thorough write-up!

1. I like how the scroll past end plugin feels.

2. I think more blank lines is a good idea. 25 `"/n"`s looks good to me.

3. I'm not sure if I prefer line-wrapping or the horizontal scroll that we had before. While line-wrapping means that you don't have to scroll to read what it says, I think that it makes it less clear what's going on in the program, though this is largely a taste thing. We can try it and see what happens when students use it. Potentially the smaller "I have a question" window that you mention in (4) solves this problem so we don't have to do line-wrapping because students can see the horizontal scroller. 

4. I like what you did here! Can you take this even further and change "I have a question ^" to "Chat"? (Another idea, which might be tricky, is to put the button to make the chat show up in the navbar near the Clean Code button. That way it won't get in the way of editing at all.)

5. I agree with you that more space without the gutters looks good and adds space which is good. This unfortunately removes the line-errors that show up in the gutter when a student makes a syntax error. It also removes the feature where students can "fold" their code so they can navigate it more quickly. I think those features are worthwhile so I'd uncomment that line.

madduccino:
3. I turned off line wrapping. I think the way it wraps URLs can be especially confusing. And as you said, chat window changes should remove issues with horizontal scrolling.

4. The chat window seems best to me on the bottom, rather than on the nav. I changed the text to Chat and switched the style.

stevekrouse:
Looks great! Last question: could we get rid of the word "Chat" and have it just be a super small button with the chat icon? Or do you think that's too confusing?

madduccino:
Either one works for me. Changed it to the button with just the icon, check it out.

stevekrouse:
I like it!

PR #324, opened 2017-03-07
make key events more comprehensive and user-friendly

### LINKED ISSUE #265: delete key needs to map to DELETE

joebeachjoebeach:
fixes #265 
plus, adds flexibility to keysDown.includes() and onKeyDown() by allowing upper/lower case, and more key name possibilities

### DIFF
--- a/src/woof.es6.js
+++ b/src/woof.es6.js
@@ -277,10 +277,49 @@ function Woof({global = false, fullScreen = false, height = 500, width = 350} = 
   thisContext.mouseYSpeed = 0;
   thisContext.keysDown = [];
   
-  //modify keysDown.includes() to not be case-sensitive
+  //modify keysDown.includes() to not be case-sensitive, and to accept more user input possibilities
   thisContext.keysDown.oldIncludes = thisContext.keysDown.includes
-  thisContext.keysDown.includes = function(item) {
-    return this.oldIncludes(item.toUpperCase())
+  thisContext.keysDown.includes = function(arg) {
+    var key = arg.toUpperCase()
+    
+    if (key === "UP ARROW") {
+      key = "UP"
+    }
+    else if (key === "LEFT ARROW") {
+      key = "LEFT"
+    }
+    else if (key === "RIGHT ARROW") {
+      key = "RIGHT"
+    }
+    else if (key === "DOWN ARROW") {
+      key = "DOWN"
+    }
+    else if (key === "RETURN") {
+      key = "ENTER"
+    }
+    else if (key === "CONTROL") {
+      key = "CTRL"
+    }
+    else if (key === "OPTION") {
+      key = "ALT"
+    }
+    else if (key === "ESC") {
+      key = "ESCAPE"
+    }
+    else if (key === "SPACE BAR" || key === "SPACEBAR") {
+      key = "SPACE"
+    }
+    else if (key === "CAPS") {
+      key = "CAPS LOCK"
+    }
+    else if (key === "DEL" || "BACKSPACE") {
+      key = "DELETE"
+    }
+    else if (key === "CMD" || key === "WINDOWS" || key === "SEARCH") {
+      key = "COMMAND"
+    }
+    
+    return this.oldIncludes(key)
   }
   
   thisContext.ready(() => {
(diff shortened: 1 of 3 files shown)


stevekrouse:
View this pull request at:

https://cdn.rawgit.com/stevekrouse/WoofJS/dcac3d485282ac561ab30550c2eadba8d261dd48/index.html

src/woof.es6.js
```diff
@@ -303,17 +310,35 @@ function Woof({global = false, fullScreen = false, height = 500, width = 350} =
...
       event.preventDefault();
     });
-    
+
     document.body.addEventListener("keydown", event => {
       var key = Woof.prototype.keyCodeToString(event.keyCode);
-      if (!thisContext.keysDown.includes(key)){
-       thisContext.keysDown.push(key); 
-      }
+        if (typeof key == "object") { //if keyCodeToString returns an array, eg ["SPACE", "SPACE BAR", "SPACEBAR"]
```
stevekrouse: We normally put comments above the line they are referring to. We also normally put a space between `//` and the next word.

src/woof.es6.js
```diff
@@ -276,6 +276,13 @@ function Woof({global = false, fullScreen = false, height = 500, width = 350} =
   thisContext.mouseXSpeed = 0;
   thisContext.mouseYSpeed = 0;
   thisContext.keysDown = [];
+  
+  //modify keysDown.includes() to not be case-sensitive
+  thisContext.keysDown.oldIncludes = thisContext.keysDown.includes
+  thisContext.keysDown.includes = function(item) {
```
stevekrouse: I see that your goal in this pull request is to make the `includes()` function accept user input in a much more flexible way. I think that's a wonderful idea and will eliminate a whole class of annoying bugs students run into.

You go after this goal with two strategies:

1) You have `keysDown` include aliases, like `["SPACE", "SPACE BAR", "SPACEBAR"]`.

2) You modify the `includes()` function to recognize the lowercase versions of letters.

I like strategy (2) a lot and I think you can extend it to accomplish what strategy (1) accomplishes in a cleaner way because...

1) it keeps the keeps the `keysDown` array cleaner. You can do something like `keysDown.length` and get an accurate count. Or you can put keysDown on the screen with `new Text({text: () => keysDown})` and not see duplicates.

2) It allows you to keep the logic of aliases all in one place. You wouldn't need `if (typeof key == "object")` all over your code. 

I would make `keyCodeToString` go back to only returning strings (and not arrays of strings) so that the interface to keysDown is the same everywhere. 

Then in the `includes()` method, I would check `if (item.toUpperCase() == "SPACE BAR" || item.toUpperCase() == "SPACEBAR"` and map it to "SPACE" there.

src/woof.es6.js
```diff
@@ -332,15 +357,36 @@ function Woof({global = false, fullScreen = false, height = 500, width = 350} =
...
-  
+    
     thisContext._onKeyDownHandler = event => {
       var key = Woof.prototype.keyCodeToString(event.keyCode);
-      thisContext._onKeyDowns.forEach((func) => {func(key)});
+      
+      if (typeof key == "object") { // if keyCodeToString returns an array e.g. ["SPACE", "SPACE BAR", "SPACEBAR"]
+        key.forEach(item => {
+          thisContext._onKeyDowns.forEach((func) => {func(item)});
+          thisContext._onKeyDowns.forEach((func) => {func(item.toLowerCase())}); // eliminate case-sensitivity
```
stevekrouse: I might be missing something, but it seems like this would call each `onKeyDown` handler 2-6x as many times as it should. 

For example if the space bar is pressed, it would call each onKeyDown handler for `["SPACE", "SPACE BAR", "SPACEBAR"]` and  `["space", "space bar", "spacebar"]`.

stevekrouse: (I understand that you're trying to make the key that is fed into key event handlers to be case-insensitive. I don't think it's possible, however. The function can only be run once and it can only be fed one key, so you can't give it both. On the bright side, I don't think many kids use the key event handler's `key` parameter very much, so it should be alright.)

stevekrouse: Here are some ideas we can do to mitigate this...

1. We could change the documentation from

```javascript
onKeyDown(key => {
 if (key == 'A') {
   /* do something here */
  }
})
```
to 
```javascript
onKeyDown(() => {
 if (keysDown.includes('a')) {
   /* do something here */
  }
})
```
Basically, we don't change how the function works, but we almost pretend like it doesn't work that way.

2. We could add a new function to woof
```javascript
onKeyDown('a', () =>
  /* do something here */
})
```
This function we could make case-insensitive and work with aliases.

PR #334, opened 2017-03-17
made a better error message for broken link

nicolekelner:


stevekrouse:
View this pull request at:

https://cdn.rawgit.com/stevekrouse/WoofJS/223ef36a999395025689d1bf006be47048de76f2/index.html

create.html
```diff
@@ -310,6 +310,10 @@
                 editor.setValue(editor.getValue().replace(/sprite1/g, spriteName));
             }  
           }
+          
+          if (this.error.message.includes('CanvasRenderingContext2D') && $('#logggedin').is(':visible')) {
```
stevekrouse: Given that errors show up for users whether they are logged in or not, I don't see any reason why we should only show users the more understandable error if `$('#logggedin').is(':visible')`. (I know we did this last time because we didn't want the alert to pop up constantly if the user wasn't logged in. However with the design where the message only pops up when you click the error message, we don't need this.) Please remove ` && $('#logggedin').is(':visible')` here and on line 307. 

The way you are detecting this error by seeing if the message includes `'CanvasRenderingContext2D'` is pretty good but you could be more precise by including the whole error message... `"Uncaught InvalidStateError: Failed to execute 'drawImage' on 'CanvasRenderingContext2D': The provided 'HTMLImageElement' was provided in a broken state."` The same goes for the above error that only checks that the message includes "sprite1". Being precise here is important because we only want to specifically catch one type of error and not just any error that might have similar key words to the error that we want to catch.

To be clear, I asked for both of these changes  in the last pull request, but they must've gotten lost in the shuffle: https://github.com/stevekrouse/WoofJS/pull/321#discussion-diff-105480537R338

create.html
```diff
@@ -310,6 +310,10 @@
...
+           alert("You're using a broken link to an image. Take a look at the link to your image and make sure it is correct!");
```
stevekrouse: It looks to me like this line is indented with one space under the if-statement. Please ensure that it is two spaces.

Students often have a number of images on the page when they get this error, so it's more a process of looking through all their images to find the broken one. Maybe you could acknowledge that they'll have to look through multiple images to find the offending one? It could also be helpful to suggest that they might want to comment out all of their images and uncomment them out one by one to find the offending image. You could also suggest that any image they recently added could be a likely suspect. Just some ideas -- I'll leave it to you to craft the message however you see fit.

create.html
```diff
@@ -304,12 +304,16 @@
...
         getID: getID,
         showError: function(){
-          if (this.error.message.includes('sprite1') && $('#logggedin').is(':visible')) {
+          if (this.error.message.includes('Uncaught ReferenceError: sprite1 is not defined')) {
             var spriteName = prompt("It looks like you have not renamed your your sprite! It is still set as the default 'sprite1'.\n\nWhat is the name of your sprite?", "unicornSprite");    
             if (spriteName !== null) {   
                 editor.setValue(editor.getValue().replace(/sprite1/g, spriteName));
             }  
           }
+
```
stevekrouse: Please delete this empty line :)

PR #335, opened 2017-03-19
make random return same number of decimals

### LINKED ISSUE #332: random() should give you back as many decimals as you give it
```javascript
random(1, 10)
=> 5
random(1.2, 10)
=> 5.7
random(1.35, 10)
=> 5.89
```

joebeachjoebeach:
`random()` now returns a number with the same number of decimals as the numbers you put in

closes #332

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -869,14 +869,6 @@
                             code: "random(1, 10)",
                             tags: "random"
                         }, {
-                            description: "You can control how many decimals the random number has",
-                            code: "random(2.22, 10)\n\/\/ 8.51\nrandom(1.1234, 2.6789)\n\/\/ 2.1097\nrandom(2, 5)\n\/\/ 4",
-                            tags: "random"
-                        },{
-                            description: "Unfortunately, JavaScript doesn't understand what you mean if you add zeros in the decimals place, so try adding a 1 at the end",
-                            code: "random(2.00, 5.00)\n\/\/ 3\nrandom(2.01, 5)\n\/\/ 4.28",
-                            tags: "random"
-                        },{
                             description: "Random X value on the screen between minX and maxX:",
                             code: "randomX()"
                         }, {
--- a/docs/index.html
+++ b/docs/index.html
@@ -869,6 +869,14 @@
                             code: "random(1, 10)",
                             tags: "random"
                         }, {
+                            description: "You can control how many decimals the random number has",
+                            code: "random(2.22, 10)\n\/\/ 8.51\nrandom(1.1234, 2.6789)\n\/\/ 2.1097\nrandom(2, 5)\n\/\/ 4",
+                            tags: "random"
+                        },{
+                            description: "Unfortunately, JavaScript doesn't understand what you mean if you add zeros in the decimals place, so try adding a 1 at the end",
+                            code: "random(2.00, 5.00)\n\/\/ 3\nrandom(2.01, 5)\n\/\/ 4.28",
+                            tags: "random"
+                        },{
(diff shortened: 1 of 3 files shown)


stevekrouse:
View this pull request at:

https://cdn.rawgit.com/stevekrouse/WoofJS/5ebd9545e630b6324fa5237707b11f7728b50bf4/index.html

src/woof.es6.js
```diff
@@ -1262,7 +1262,24 @@ Woof.prototype.random = function(a, b) {
   if (Number.isInteger(min) && Number.isInteger(max)) {
```
stevekrouse: It'd be nice if we didn't have to treat integers as a special case. Could they simple be a number with 0 decimal points?

src/woof.es6.js
```diff
@@ -1262,7 +1262,24 @@ Woof.prototype.random = function(a, b) {
...
+    }
+    if (Math.floor(max) === max) {
+      maxDecimals = 0
+    }
+    else {
+      maxDecimals = max.toString().split(".")[1].length || 0
+    }
+    
+
+    return rand.toFixed(Math.max(minDecimals, maxDecimals));
```
stevekrouse: `toFixed()` returns a string. 

Trying adding something to it, like `random(1.5, 2) + 9` and you'll get `1.X9` where X is a random number between 6 and 9.

src/woof.es6.js
```diff
@@ -1262,7 +1262,24 @@ Woof.prototype.random = function(a, b) {
   if (Number.isInteger(min) && Number.isInteger(max)) {
     return Math.round(rand);
   } else {
-    return rand;
+    
+    var minDecimals, maxDecimals
+    
+    if (Math.floor(min) === min) {
```
stevekrouse: Under what conditions is `Math.floor(min) === min` true? It's unclear to me why you have to check for this.

src/woof.es6.js
```diff
@@ -1262,7 +1262,24 @@ Woof.prototype.random = function(a, b) {
...
+      minDecimals = 0
+    } 
+    else {
+      minDecimals = min.toString().split(".")[1].length || 0
```
stevekrouse: It's not clear to me when `min.toString().split(".")[1].length` would ever be false-y and would require the `|| 0` there.

src/woof.es6.js
joebeachjoebeach: Oops. Oversight on my part -- I had that for both `min` and `max` to check if they were integers. I removed the `max` one and somehow overlooked the `min` one.

docs/index.html
```diff
@@ -869,6 +869,14 @@
                             code: "random(1, 10)",
                             tags: "random"
                         }, {
+                            description: "You can control how many decimals the random number has",
+                            code: "random(2.22, 10)\n\/\/ 8.51\nrandom(1.1234, 2.6789)\n\/\/ 2.1097\nrandom(2, 5)\n\/\/ 4",
+                            tags: "random"
+                        },{
+                            description: "Unfortunately, JavaScript doesn't understand what you mean if you add zeros in the decimals place, so try adding a 1 at the end",
```
stevekrouse: I think we could combine these three blocks into one: 

> `random()` returns as many decimals as it gets. (Note: Trailing `0`s are ignored in JavaScript. For example, `2.50` is recognized as having only one decimal, so we advise you make your trailing `0` into a `1`, because `2.51` is recognized as having two decimals.)

I also think that we can do away with the examples and their output in the code section. I try to avoid that wherever possible so as for easier copying and pasting.

PR #340, opened 2017-03-22
when refresh, focus on canvas

nicolekelner:


stevekrouse:
View this pull request at:

https://cdn.rawgit.com/stevekrouse/WoofJS/ab7e83efdf2f6e0d25a188ea8d7788f3bbc44279/index.html

create.html
```diff
@@ -555,6 +555,13 @@
       debouncedRunCode()
     });
     runCode()
+    
+    // focus on the canvas when refreshed
+     function focuscanvas() {
+       var canvas = document.getElementById('preview');
```
stevekrouse: The `#preview` element you're grabbing is an iframe, not a canvas so the variable name is misleading. It's common practice to name variable after the id, so you could call it `preview`.

create.html
```diff
@@ -555,6 +555,13 @@
...
+       canvas.setAttribute('tabindex', '0');
```
stevekrouse: I'm not sure if you need this line. I removed this line and it still worked on my computer. Does it work on yours without this line?

create.html
```diff
@@ -188,7 +188,7 @@
         <div id="PreviewNav" @click="previewbar = !previewbar" style="height:25px;text-align:center; background-color:#4875D1; color:white; font-size:1.1em style; display:flex; justify-content: space-between;">
           <span style="margin-left:8px;  margin-top:4px">Preview</span>
           <span style="margin-right:8px">
-            <img @click="previewbar = !previewbar"  onclick="runCode()" id="pop" style="margin: 2px; width: 17px; height:17px" src="./images/refresh.png">
+            <img @click="previewbar = !previewbar"  onclick="runCode(); focuscanvas()" id="pop" style="margin: 2px; width: 17px; height:17px" src="./images/refresh.png">
```
stevekrouse: Sorry I didn't notice this before, but we do camelCase in js so this should be `focusCanvas()`

create.html
```diff
@@ -555,6 +555,12 @@
       debouncedRunCode()
     });
     runCode()
+    
+    // focus on the canvas when refreshed
```
stevekrouse: "focus on the **preview**"

create.html
```diff
@@ -555,6 +555,12 @@
...
+     function focuscanvas() {
+       var preview = document.getElementById('preview');
```
stevekrouse: you can do this all in one line, without a variable

create.html
stevekrouse: (It should actually be `focusPreview()` because you're not focusing on a canvas.)

PR #344, opened 2017-03-30
add instructions for removing object from list

madduccino:


stevekrouse:
View this pull request at:

https://cdn.rawgit.com/stevekrouse/WoofJS/fa7e7161e5288f42aaaff82ca58e9f2fc93661fb/index.html

stevekrouse:
I like the idea of adding this instruction near where we explain clones. However, I don't love the idea of combining it with `.delete()` because they don't *always* go together.

Other ideas:
  * create a separate entry below delete that explains how to remove clones from a list
  * create a note in text (not in the code) above delete that explains that it only delete from the screen and to truly remove it, you may need to remove it from it's array.

Also: I think the most important place to put this is in the cloning example code with a comment https://github.com/stevekrouse/WoofJS/pull/344/files#diff-eae600270579316beb8b8d9b9a5732d4R692

madduccino:
I _think_ this picks up your comments

Issue #338, opened 2017-03-21
backdrop canvas div width=0, height=0 in jsbin, jsfiddle but not woofjs.com/create


stevekrouse:

stevekrouse:
This is where we set the width and height of the backdrop canvas div: https://github.com/stevekrouse/WoofJS/blob/master/src/woof.es6.js#L125

And this is where we get the width and height of the screen for fullscreen: https://github.com/stevekrouse/WoofJS/blob/master/src/woof.es6.js#L72

However when I debugged this by setting breakpoints, it seems liked this code was running properly. I have a strange feeling that this is a weird bug.

One difference between JSbin and Woof is that in woof we synthetically trigger a 'load' event so that woof thinks the page is being loaded. To be honest, I forget exactly why we do this but that might be part of it... https://github.com/stevekrouse/WoofJS/blob/master/create.html#L625

madduccino:
Here's where it broke: https://github.com/stevekrouse/WoofJS/commit/1e9b9a5e25ec87307efbe17924a09c4e0865fe6e

How would I go from here again?

stevekrouse:
Ah, that makes sense!

Sometimes, a woof developer forgets to commit the new compiled core library after they make changes to it. Their changes don't come into the core library until the next developer to edit the core library compiles it. That's what happened here.

So it looks like the bug came from this pull request: https://github.com/stevekrouse/WoofJS/pull/276

That's a long time ago! But it makes sense. That pull request totally changed the way we render backdrops.

Moving on... the most pressing question in my mind is: why does this work on woofjs.com/create and not on jsbin or jsfiddle? It feels important to understand that before we move on. My suspicion is that it has something to do with how to set the CSS of the iframe on /create.

You could try to reproduce this bug on cloud 9 by creating an HTML page, including Woof, and previewing it. I'd be curious if the bug shows up there. 

Let me know if you'd like more debugging ideas! This is a hard one so at the very least you'll get to see how my mind works when I'm debugging.

madduccino:
If you change "div" back to "canvas" on this line, the bug is resolved: 

https://github.com/stevekrouse/WoofJS/blob/master/src/woof.es6.js#L122

Also, if you remove the Doctype declaration, the bug doesn't happen either.

So it has something to do with how the DOM is treating canvas vs div.

This makes the background blue:

[...]

PR #346, opened 2017-04-02
make backdrop div 100% width/height

madduccino:
#338 - reproduced the bug in /create by adding doctype. Added 100% width/height to html and body tags and #project and #backdrop divs which resolved it.

### DIFF
--- a/create.html
+++ b/create.html
@@ -615,9 +615,8 @@
         setTimeout(function() {
         
             // add doctype to iframe
-            var myContent = '<!DOCTYPE html>'
-            iframe.contentWindow.document.open('text/htmlreplace');
-            iframe.contentWindow.document.write(myContent);
+            iframe.contentWindow.document.open();
+            iframe.contentWindow.document.write('<!DOCTYPE html>');
             iframe.contentWindow.document.close();
             
             // add a base tag to the page so it knows where to pull relative image urls
--- a/create.html
+++ b/create.html
@@ -613,6 +613,13 @@
         
         iframe.src = ""; // first, we clear the preview iframe
         setTimeout(function() {
+        
+            // add doctype to iframe
+            var myContent = '<!DOCTYPE html>'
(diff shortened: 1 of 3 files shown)


stevekrouse:
View this pull request at:

https://cdn.rawgit.com/stevekrouse/WoofJS/a7e491626756353c5830b2576876070f63174ac9/index.html

src/woof.es6.js
```diff
@@ -95,11 +95,17 @@ function Woof({global = false, fullScreen = false, height = 500, width = 350} =
   
   window.addEventListener("load", () => {
     // create the main div that Woof lives in
```
stevekrouse: This comment is now out of place. Can you put it where it belongs, above `thisContext._mainDiv = document.createElement("div");`?

src/woof.es6.js
```diff
@@ -95,11 +95,17 @@ function Woof({global = false, fullScreen = false, height = 500, width = 350} =
...
+    document.documentElement.style.width = "100%";
+    document.documentElement.style.height = "100%";
```
stevekrouse: I've never seen `document.documentElement` before. Do you need it here? It seems like the `document.body` code below should take care of this, no?

create.html
```diff
@@ -613,6 +613,13 @@
         
         iframe.src = ""; // first, we clear the preview iframe
         setTimeout(function() {
+        
+            // add doctype to iframe
+            var myContent = '<!DOCTYPE html>'
+            iframe.contentWindow.document.open('text/htmlreplace');
```
stevekrouse: I can't find anywhere online to support `text/htmlreplace`. It seems like you can leave it out as it says here https://www.w3schools.com/jsref/met_doc_open.asp

src/woof.es6.js
madduccino: document.documentElement adds style to the html tag which is necessary - 100% width/height has to be set on html and body.

Issue #323, opened 2017-03-06
Add Audio to WoofJS core library


stevekrouse:
It seems like we can wrap the JS Audio library a bit to make it more intuitive. For starters, we can simply add a few methods to the JS Audio prototype...

```javascript
Audio.prototype.startPlaying = function() { ... }
```

It could work like so...

```javascript
var sound = new Audtio('some URL')

// makes it clear that it goes to the next line before the playing is stopped 
//unlike in scratch when you can "play until done"  
sound.startPlaying() 

 // more clear then sound.played.length === 0
if (sound.neverPlayed) { ... } 
```

For reference... https://www.w3schools.com/jsref/dom_obj_audio.asp

PR #347, opened 2017-04-02
add audio methods

madduccino:
#323 

Question: How are these images files created?

url: "./images/68747470733a2f2f692e696d6775722e636f6d2f4c387a6e6d4f622e706e67",


stevekrouse:
View this pull request at:

https://cdn.rawgit.com/stevekrouse/WoofJS/b1b25dbc6701317538f1a006ae5c36d127763317/index.html

dist/woof.js
```diff
@@ -1894,3 +1894,31 @@ if (JSON.parse(currentScript.getAttribute('global')) !== false) {
...
+  this.loop = true;
+};
+
+Audio.prototype.neverPlayed = function () {
+  if (this.played.length === 0) {
+    return true;
+  }
+};
+
+Audio.prototype.changeSpeed = function (speed) {
```
stevekrouse: How about `setSpeed`?

stevekrouse: How about `setSpeed`?

dist/woof.js
```diff
@@ -1894,3 +1894,31 @@ if (JSON.parse(currentScript.getAttribute('global')) !== false) {
...
+  this.playbackRate = speed;
+};
+
+Audio.prototype.changeVolume = function (vol) {
```
stevekrouse: And how about `setVolume`?

dist/woof.js
```diff
@@ -1894,3 +1894,31 @@ if (JSON.parse(currentScript.getAttribute('global')) !== false) {
   // unless the script tag containing Woof has an attribute global="false", start Woof in global mode
   Woof.prototype.extend(window, new Woof({ global: true, fullScreen: true }));
 }
+
+// Add methods to JS audio prototype
+Audio.prototype.startPlaying = function () {
```
stevekrouse: I wonder if we should also wrap the creation of an `Audio` object. When you create a sprite, you can set everything all at once as you create it. Maybe you should be able to set looping, speed and volume upon creation as well:

```javascript
var bruce = new Audio({
  url: 'http://mhoerner.dyndns.org/mp3/Bruce%20Springsteen/Born%20To%20Run/05%20Born%20To%20Run.mp3',
  loop: true,
  volume: 1,
  speed: 3
})
```

And maybe you should be able to modify those things in the same way as with sprites:

```
bruce.speed = 5
bruce.volume = 1
```

madduccino: I'll get to work on this. Regarding my image question above, I was wondering what file format this is: ./images/68747470733a2f2f692e696d6775722e636f6d2f4c387a6e6d4f622e706e67. Why is there no .jpg/png/gif extension etc?

stevekrouse: Oh, that's a mistake. Sometimes things work without extension but there should be one.

dist/woof.js
stevekrouse: Or just `.speed = `

dist/woof.js
stevekrouse: Or just `.volume = `

madduccino:
I can't figure out how to wrap the Audio object. I see where the sprite prototypes are created in the core library but can't put it all together.

stevekrouse:
Good question! What I'm talking about here is also referred to as "inheritance." You can read more about it all over the internet but this seems like a decent place to start: 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Inheritance_and_the_prototype_chain

Let me know if you'd like more help :)

madduccino:
I finally got this to work!! Hooray. I also added play/pause buttons so users can preview the sound before they place it in their code. This has been requested by a few students.

src/woof.es6.js
```diff
@@ -1245,6 +1245,111 @@ Woof.prototype.Image = function({project = undefined, url = "./images/SMJjVCL.pn
   };
 };
 
+Woof.prototype.Sound = function({project = undefined, url = '', loop = false, speed = "normal", volume = "normal"} = {}) {
```
stevekrouse: Do you need `project = undefined` here?

src/woof.es6.js
```diff
@@ -1245,6 +1245,111 @@ Woof.prototype.Image = function({project = undefined, url = "./images/SMJjVCL.pn
...
+  this.url = url;
```
stevekrouse: Do you need to save the URL to `this`?

src/woof.es6.js
```diff
@@ -1245,6 +1245,111 @@ Woof.prototype.Image = function({project = undefined, url = "./images/SMJjVCL.pn
...
+  // Create new audio object with given url
+  var sound = new Audio(url);
+  // Define loop, speed, and volume properties
```
stevekrouse: Line 1252 to line 1255 seem unnecessary. Can't you just set these properties directly on the `sound` object?

src/woof.es6.js
```diff
@@ -1245,6 +1245,111 @@ Woof.prototype.Image = function({project = undefined, url = "./images/SMJjVCL.pn
...
+  
+  // These functions are wrappers that allow the native audio functions to execute
+  setLoop(loop);
+  setSpeed(speed);
+  setVolume(volume);
+  
+  // Allow user to get and set speed
+  Object.defineProperty(this, 'speed', {
+    get: function() {
+      return this.soundSpeed;
```
stevekrouse: Can't you return `sound.speed`?

src/woof.es6.js
```diff
@@ -1245,6 +1245,111 @@ Woof.prototype.Image = function({project = undefined, url = "./images/SMJjVCL.pn
...
+    },
+    set: function(value) {
+    if (value != "slow" && value != "fast" && value != "normal") {
+      throw new Error("Speed can only be set to normal, slow, or fast.");
```
stevekrouse: Can't you allow those three values and also any number?

src/woof.es6.js
stevekrouse: Same goes for the other properties

src/woof.es6.js
stevekrouse: Same goes for volume

src/woof.es6.js
```diff
@@ -1245,6 +1245,111 @@ Woof.prototype.Image = function({project = undefined, url = "./images/SMJjVCL.pn
...
+    }
+  });
+  
+  // Allow user to get and set loop
+  Object.defineProperty(this, 'loop', {
+    get: function() {
+      return this.soundLoop;
+    },
+    set: function(value) {
+      if (typeof value != "boolean") {
```
stevekrouse: Good type check!

src/woof.es6.js
```diff
@@ -1245,6 +1245,111 @@ Woof.prototype.Image = function({project = undefined, url = "./images/SMJjVCL.pn
...
+  };
+  
+  // Stop the sound
+  this.stopPlaying = function() {
+    sound.pause();
+  };
+  
+  // Check if the sound has been played
+  this.neverPlayed = function() {
+    if (sound.played.length === 0) {
```
stevekrouse: This function simplifies to

```javascript
return sound.played.length === 0
```

src/woof.es6.js
```diff
@@ -1245,6 +1245,111 @@ Woof.prototype.Image = function({project = undefined, url = "./images/SMJjVCL.pn
...
+    }
+  });
+  
+  // Set loop to true or false
+  function setLoop(val) {
+    sound.loop = val ? true: false;
+  }
+
+  // Set speed to slow, normal, or fast
+  function setSpeed(val) {
```
stevekrouse: I would change this function to `convertSpeedToNumber`, and it would either convert the string to a number or it would just pass the number it was given directly through (as I described above)

src/woof.es6.js
```diff
@@ -1245,6 +1245,111 @@ Woof.prototype.Image = function({project = undefined, url = "./images/SMJjVCL.pn
...
+        throw new Error("Loop must be set to true or false."); 
+      }
+      this.soundLoop = value;
+      setLoop(value);
+    }
+  });
+  
+  // Set loop to true or false
+  function setLoop(val) {
+    sound.loop = val ? true: false;
```
stevekrouse: This simplifies to 

```javascript
sound.loop = val
```

docs/index.html
```diff
@@ -51,6 +51,10 @@
               <a v-if="block.integrationDocumentationLink" target='_blank' class="btn btn-primary" :href="block.integrationDocumentationLink" role="button" style="width:100%;">Full Documentation</a>
             </div>
             <img v-if="block.url" ondragstart="event.dataTransfer.setData('text/plain', event.target.alt)" :alt="block.code" :src="block.url">
+            <div v-if="block.sound" class="play-sound">
+              <span @click="playSound(block.sound)" class="glyphicon glyphicon-play-circle" aria-hidden="true"></span>
```
stevekrouse: I wouldn't spend too much time on it, but it would be neat if it only showed pause when the sound was playing, and vice versa.

docs/docs.css
```diff
@@ -502,3 +502,14 @@ h4 {
     background-color: #563d7c;
     border-color: #563d7c;
 }
+
+.play-sound {
+  display: inline;
```
stevekrouse: I think it'd be good to add 

```css
cursor: pointer;
```

to this one and the pause one as well so it's more clear they are buttons.

docs/index.html
```diff
@@ -1380,126 +1394,102 @@
                     },
                     ],
                     Sound: [{
+                        sound: "./sounds/mario-jump.wav",
                         url: "./images/playmario.png",
-                        code: "var mario = new Audio('./docs/sounds/mario-jump.wav')\nmario.play()",
+                        code: 'var mario = new Sound({\n  url: "./docs/sounds/mario-jump.wav",\n  loop: false, // options: true, false \n  speed: "normal", // options: fast, slow, normal\n  volume: "normal" // options: mute, low, normal\n})\n\nmario.startPlaying()',
```
stevekrouse: It's kinda nice to give all of the options, but it gets crowded, especially when they have a lot of other things on the screen or have Woof in half-screen mode.

I would suggest moving the options to a section at the bottom similar to how we have it for textAlign:

screenshot 2017-10-25 at 1 20 08 pm

src/woof.es6.js
madduccino: I just copied that from another Woof.prototype declaration. What does it mean?

stevekrouse: Woof sprite objects need to know what project they are in so they can be appropriately rendered to that specific project. (for example, imagine multiple seperate woof projects on the same page.) If the project is in global mode, it just makes the project the window object. https://github.com/stevekrouse/WoofJS/blob/master/src/woof.es6.js#L557

Woof Sound is just a helper object. It doesn't need to be tied to a Woof project in order to run

nicolekelner:
slice 6 1
Here's the play button png

madduccino:
@nicolekelner thanks for this totally rad play button. Can you make me a pause SVG? 🙏

madduccino:
Or a stop button. I used pause because I thought it looked better, but maybe a square will with the way you're making it.

madduccino:
-Made improvements to Woof.prototype.sound definition
-Made play button "turn into" pause when playing and back to play when sound finishes playing
-Allowed for both string and number inputs to volume and speed and included these options at the bottom as opposed to inside the code

docs/docs.css
stevekrouse: @madduccino just wanted to make sure you saw this

src/woof.es6.js
```diff
@@ -1245,6 +1244,91 @@ Woof.prototype.Image = function({project = undefined, url = "./images/SMJjVCL.pn
...
 };
 
+Woof.prototype.Sound = function({url = '', loop = "false", volume = "normal", speed = "normal"} = {}) {
+  
+  // Create new audio object with given url
+  this.audio = new Audio(url);
+  this.audio.loop = loop;
+
+  // Store allowed audio speed and volume values
+  this.audioSettings = {
```
stevekrouse: This looks great. A few minor points:

1. Can you make `Speed` and `Volume` lower case? That's the convention for JS Object property names.
2. You shouldn't need to duplicate data. For example `slow: [0.5, "slow"]` has `slow` twice. Can you do it with only `slow` once?

src/woof.es6.js
```diff
@@ -1245,6 +1244,91 @@ Woof.prototype.Image = function({project = undefined, url = "./images/SMJjVCL.pn
...
+      low: [0.5, "low"],
+      normal: [1, "normal"]
+    }
+  }
+  
+  // Convert given value to corresponding audio object value
+  // Throw error if given value not found in allowed values
+  function convertAudioSettings(propName, prop, val) {
+    var props = [];
+    for (var key in prop) {
```
stevekrouse: I have trouble figuring out how this function works... Thoughts:

1. It's a bit confusing to have so many variables with similar names: `propName`, `key` and `prop`. Maybe think about more clear names.
2. Fixing the duplication of data thing above that I mentioned might help this become simpler.
3. If a student wants to set `sound.speed = 0.3` will that be allowed?

madduccino: This function should be easier to follow now. If a student provides an unacceptable value, an error will be thrown. I tried really hard to not have to pass the third argument, originally called `propName` and now called `objName` but I couldn't find another way to include it in the error message. Perhaps this reveals a structural problem in the `this.audioSettings` object? Basically, I wanted to be able to grab the property `speed` which is also an object but the closest I came was `Object.keys(this.audioSettings)` which returns `["speed", "volume"]` From there, I could get `speed` from `Object.keys(this.audioSettings)[0]`, but I couldn't find a dynamic way of getting the 0 index.

docs/docs.css
madduccino: I did  - cursor is set to pointer.

src/woof.es6.js
```diff
@@ -1245,6 +1244,92 @@ Woof.prototype.Image = function({project = undefined, url = "./images/SMJjVCL.pn
...
+    volume: {
+      mute: 0,
+      low: 0.5,
+      normal: 1
+    }
+  }
+  
+  // Convert given value to corresponding audio object value
+  // Throw error if given value not found in allowed values
+  function convertAudioSettings(obj, val, objName) {
```
stevekrouse: We may be trying to be too clever here. It's still difficult for me to follow this function. Why don't we make it super simple with two functions? The following is what I'm thinking for one of them:

```javascript
const soundVolumeToAudioVolume = val => {
  if (typeof val == "number") {
    if (val >= 0 && val <= 1) {
      return val 
    } else {
      throw Error("...")
    }
  } else if (typeof val == "string") {
    if (val == "slow") {}
    else if (val == "normal") {}
    else if (val == "fast") {}
  } else {
    throw Error("..")
  }
}
```

This way we can clearly see all of the possible inputs and outputs.

stevekrouse:
Looks great! I made one last comment on the audio conversion function. When that's done, feel free to merge in yourself :)

Issue #175, opened 2016-09-22
every in woof could take a function instead of number so they we could have random intervals


stevekrouse:

stevekrouse:
for example, if you want clones to show up every random number of seconds between 2 and 7 seconds, you really can't do that now. 

I'm proposing making, `every(() => random(2,7), "seconds", () => {...})` a valid way to do that.

Issue #304, opened 2017-02-24
remove the ability to set text to functional expression


stevekrouse:
This is a change for the simpler, but it would make things harder and clunkier.

This would also be a breaking change which would be annoying to build out on the backend so that old code works on the previous version and new code works on the newer version. We'd also have to update the documentation. 

(Of course, we could make everything reactive and be able to be set to a function, which would be the opposite of this. That could also make things easier to understand.)

Issue #133, opened 2016-06-20
allow all values to be set to functions


stevekrouse:
The problem is what happens if someone then forgets and tries to add 1 to their x value that they set to a function?

Should we try to catch this error for them? If they set it to a new value, let it happen, if they try to change it by 1, we should run it, and then add 1, and then store it back in itself as a number.

stevekrouse:
This is really just a shorthand for a forever so we probably don't need it anywhere but for text because that use case is so pervasive

PR #351, opened 2017-04-17
allow every to use random intervals

joebeachjoebeach:
I don't see this as being finished, but I want feedback first.

The way I see it, we can't do `every(random(1, 10), 'seconds', () => {...})` because `random()` will just evaluate to a single number, so we need a way to run `random()` within the function itself.

My idea was to allow the `time` parameter passed to `every()` be an array, so you could do something like this: `every([random, 1, 10], 'seconds', () => {...})`

Also, in order for the action to reoccur at random intervals, we can't use `setInterval()`, because that's for regular intervals. So I think we have to repeatedly generate a random number, then use `setTimeout()` with that number, etc. etc.

adding edit: addresses #175  

Another possibility would be to branch this off into a new function called `everyRandom()` -- then you could use it like this: `everyRandom(1, 10, 'seconds', () => {...})`

### DIFF
--- a/src/woof.es6.js
+++ b/src/woof.es6.js
@@ -435,27 +435,42 @@ function Woof({global = false, fullScreen = false, height = 500, width = 350} = 
   };
 
   thisContext._everys = [];
+  // thisContext.every = (time, units, func) => {
+  //   if (typeof func != "function" || (typeof time != "number" && typeof time != "function")) { throw new TypeError("every(time, units, function) requires a number/function, time unit and function input."); }
+  //   // if the user inputs something like () => random(1, 10) for the time parameter, re-evaluate the function every time it's run, and update the frequency
+  //   if (typeof time == "function") {
+  //     if (typeof time() != "number") { throw new TypeError("every(time, units, function) requires a time function that returns a number")}
+  //     // var init = function() {
+  //     //   var theFunction = function() {
+  //     //     var ms = Woof.prototype.unitsToMiliseconds(time(), units);
+  //     //     func();
+  //     //     // use setTimeout() here instead of setInterval() because the interval has to be able to change
+  //     //     thisContext._everys.push(setTimeout(theFunction, ms));
+  //     //   }
+  //     //   theFunction();
+  //     // }
+  //     // thisContext._everys.push(init());
+  //     // init();
+  //     // var theFunction = function() {
+  //     //   var ms = Woof.prototype.unitsToMiliseconds(time(), units);
+  //     //   func();
+  //     //   // use setTimeout() here instead of setInterval() because the interval has to be able to change
+  //     //   thisContext._everys.push(setTimeout(theFunction, ms));
+  //     // }
+  //     // theFunction();
+  //   }
+  //   else {
+  //     var milis = Woof.prototype.unitsToMiliseconds(time, units);
+  //     func();
+  //     thisContext._everys.push(setInterval(func, milis));
+  //   }
+  // };
+  
   thisContext.every = (time, units, func) => {
-    if (typeof func != "function" || (typeof time != "number" && typeof time != "function")) { throw new TypeError("every(time, units, function) requires a number/function, time unit and function input."); }
-    // if the user inputs something like () => random(1, 10) for the time parameter, re-evaluate the function every time it's run, and update the frequency
-    if (typeof time == "function") {
-      if (typeof time() != "number") { throw new TypeError("every(time, units, function) requires a time function that returns a number")}
-      var init = function() {
-        var theFunction = function() {
-          var ms = Woof.prototype.unitsToMiliseconds(time(), units);
-          func();
-          // use setTimeout() here instead of setInterval() because the interval has to be able to change
-          setTimeout(theFunction, ms);
-        }
-        theFunction();
-      }
-      thisContext._everys.push(init());
-    }
-    else {
-      var milis = Woof.prototype.unitsToMiliseconds(time, units);
-      func();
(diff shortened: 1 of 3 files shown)


stevekrouse:

https://cdn.rawgit.com/stevekrouse/WoofJS/845e3b559a9f0dd7719b4b49178a339ae8abf2c4/index.html

stevekrouse:
This is very creative! I've never seen someone pass in a function as a label before!

Firstly, given that `random` here is a label, I would probably use it as a string `"random"`.

Secondly, I think a more cohesive approach would be to create a new function, `everyRandom(min, max, func)` so you didn't need to pass a label in at all.

While it does allow for the most common use-case, `everyRandom` would still make it hard to do some kinds of `every`s. For example, make a new attacker `every` `10 - points`. 

My suggestion in https://github.com/stevekrouse/WoofJS/issues/175 would solve for this use case and takes the same approach that we use for auto-updating text `() => random(3, 10 - points)`, which I'm both considering adding to everything (https://github.com/stevekrouse/WoofJS/issues/133) and removing entirely (https://github.com/stevekrouse/WoofJS/issues/304). 

What are your thoughts on all of this?

joebeachjoebeach:
I agree about the `everyRandom` idea.

I actually misread the comment on #175 and saw it as `every(random(2, 7), 'seconds', () => {...})` , but now I see better what your intention was.

In terms of the conflict between #133 and #304 -- I tend to think the best course of action is keeping things the way they are, just adding `every()` into the mix. 

I think it's overkill to make it universal, because I agree with your comment in #133 that it's really only necessary for text objects, as it's such a common use case. I would just modify that to say that it's necessary for text objects _and_ `every()`. Plus, I've seen students use `forever()` for other use cases, like setting the x position of a sprite.

I'm not sure if I see the benefit in removing this functionality entirely. I don't feel that it'd actually be a change for the simpler, as it would complicate using the framework, removing a convenient and intuitive functionality. I may be missing perspective on this, though, so I'm curious to hear where the suggestion came from.

joebeachjoebeach:
This latest commit addresses the issues above. `every()` can accept any function in the `time` parameter, so long as the function returns a number.

This means it can handle `random()` or something like `every(() => 10 - points, 'seconds', {...})` etc.

The one problem is that, in the instance of something like `every(() => 10 - points, 'seconds', {...})`, it likely wouldn't re-evaluate the function fast enough. For example, if I start with 0 points, but get 5 points within the first several seconds, it will still be waiting for the first 10 seconds before updating the frequency relative to my score.

I'm not sure how to address this. It seems to me that if we have it constantly re-evaluate, that would make the score example work better, but it would make using `random()` work terribly, because `random()` would almost always evaluate to a new number each time it's called.

Any guidance? Or is it fine how it is?

src/woof.es6.js
```diff
@@ -436,11 +436,28 @@ function Woof({global = false, fullScreen = false, height = 500, width = 350} =
...
+      var init = function() {
+        var theFunction = function() {
+          var ms = Woof.prototype.unitsToMiliseconds(time(), units);
+          func();
+          // use setTimeout() here instead of setInterval() because the interval has to be able to change
+          setTimeout(theFunction, ms);
+        }
+        theFunction();
+      }
+      thisContext._everys.push(init());
```
stevekrouse: The purpose of the `_everys` array is to keep track of all of the intervals that are running so we can cancel them if we need to. `setInterval` and `setTimeout` both return ids that allow you to cancel them if you want to. My apologies for not putting better comments in the code to this effect.

However, the init function that you define returns nothing so it doesn't make sense to add it to the `_everys` array. You probably don't need an `init` function here. Just one recursive functions should do the job. 

If we want to do this right, you'll need to add the value of `setTimeout` to the `_everys` array and also give that value to `theFunction` so that when it runs next, it can remove the prior value of `setTimeout` (because it already happened) and add the next value of `setTimeout`. Let me know if you want me to try to explain that again -- it's not easy to articulate.

stevekrouse:
This looks great! There a small technical issue I brought your attention to, but it's perfect besides this.

I don't see the issue you bring up above about re-evaluating the function fast enough as a problem. If you want something to re-evaluate constantly, you'll have to use a forever.

joebeachjoebeach:
Take a look now. All the testing I did leads me to believe that it's working properly. 

It was hard for me to figure out how to both push the value of `setTimeout()` to the `._everys` array _and_ pass it to `theFunction()` without running `setTimeout()` twice, so I pushed it to the array, then immediately pointed the `lastTimeout` variable to the last item in the array. Not sure if there's another (better?) way I could have done it.

stevekrouse:
Great, looking good!

It sounds like you want to *remember* the return value of `setTimeout` so you can use it in more than one place. It sounds like you need a...?

src/woof.es6.js
```diff
@@ -436,11 +436,44 @@ function Woof({global = false, fullScreen = false, height = 500, width = 350} =
...
+          thisContext._everys.splice(thisContext._everys.indexOf(timeoutValue), 1)
+        }
+        // use setTimeout() here instead of setInterval() because the interval has to be able to change
+        // pass an anonymous function so we can pass the argument lastTimeout to theFunction()
+        var lastTimeout = setTimeout(function() {
+          theFunction(lastTimeout)
+        }, ms)
+        
+        thisContext._everys.push(lastTimeout);
+        console.log(thisContext._everys)
```
stevekrouse: remove this console.log

stevekrouse:
Besides the `console.log`, this looks great! Go head and merge when ready 💥

Issue #221, opened 2016-10-28
make lines a thin wrapper around rectangles for collider


stevekrouse:

stevekrouse:
Lines are really just rectangles that are specified by their endpoints. Can we make the lines just a thin helper function wrapper around rectangle sprites?

PR #376, opened 2017-05-25
rectify collision detection on lines

madduccino:
#221 - 'Lines' drawn as rectangles and rotated. Further tested on Fly Trap game: http://bit.ly/2qg509N

### DIFF
--- a/dist/woof.js
+++ b/dist/woof.js
@@ -1,13 +1,13 @@
 "use strict";
 
-var _typeof = typeof Symbol === "function" && typeof Symbol.iterator === "symbol" ? function (obj) { return typeof obj; } : function (obj) { return obj && typeof Symbol === "function" && obj.constructor === Symbol ? "symbol" : typeof obj; };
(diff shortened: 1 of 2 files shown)


stevekrouse:

https://cdn.rawgit.com/stevekrouse/WoofJS/b7a5dd649323a93adc8b2dc08a9b4f85a9e4e754/index.html

madduccino:
As it turns out, the user actually can't rotate the line. But I think that's fine because they don't need to, right? Gave the comments a shot.

stevekrouse:
Yes, I think you're correct... I can imagine a scenario where a user would want to create a line (or rectangle) by one endpoint and an angle and a length, but we don't have to deal with that here.

Can you make it throw an error when the user tries to set the angle of a line sprite? You can see an example of this when the user tries to set the line height https://github.com/stevekrouse/WoofJS/pull/376/files#diff-5338d6ced086475e03af0133a2630e99R1088

madduccino:
@stevekrouse this one should be ready to merge? I'd love to have it by tomorrow as I have a student working on the fly trap game!

Issue #395, opened 2017-06-05
Line angle is off by 90 degrees


joebeachjoebeach:
Example here: http://woofjs.com/create.html#line-test

I think the rectangle and the line should be oriented the same way, but they're off by 90 degrees.

@madduccino I know you worked on `Line()` recently; can you point me in the right direction?

joebeachjoebeach:
Oh, also, I noticed `Line()`s have a `height` property, but do you think we should call it `length` instead?

stevekrouse:
Good idea, length makes more sense than height for lines

joebeachjoebeach:
OK, I'm figuring it out but still working on it. It's because, for a rectangle, `width` refers to the x dimension and `height` refers to the y dimension, but it's essentially the reverse for the line. So I'm just going to swap them then add a property `length` which will basically point to `width`

madduccino:
Sounds good Joe!

madduccino:
I wonder if it has something to do with the way I rotated the rectangle?

madduccino:
Also will your solution affect the line's collision detection? https://github.com/stevekrouse/WoofJS/blob/master/src/woof.es6.js#L629

joebeachjoebeach:
I'll make sure it doesn't mess with collision. Thanks for pointing it out!

Yeah, I think the way the line is made is basically a 90-degree rotation from the way the rectangles are made, but it shouldn't be a tough fix. I'll just have to make sure it doesn't interfere with anything else

PR #397, opened 2017-06-12
make line angle point toward endpoints

joebeachjoebeach:
Addresses #395 

@madduccino @stevekrouse can you take a look?

I swapped "width" and "height" for "length" and "thickness", since they seem to be more intuitive when dealing with a line. I'm not crazy about "thickness" but I couldn't think of a better alternative. Any thoughts?

That being said, I actually kept "width" and "height" as properties of a line object because those properties are used for rendering and for the collider. I figured it was better this way than putting more if-statements in the renderer and collider. I simply pointed `line.width` to `line.length` and `line.height` to `line.thickness` with getters, and made the setters return errors.

I still need to edit the comments and update the documentation, but I wanted to test it out here first (cloud9 is being a pain...)

### DIFF
--- a/dist/woof.js
+++ b/dist/woof.js
@@ -1043,15 +1043,9 @@ Woof.prototype.Sprite = function () {
     if (typeof steps != "number") {
       throw new TypeError("move(steps) requires one number input.");
     }
-    // moving lines with this method is tricky, so throw an error (should probably be fixed at some point)
-    if (this.type == "line") {
-      throw new TypeError("You cannot move lines with move() unfortunately! Change the x, y, x1, and y1 values instead.");
-    }
     // we modify privateX and privateY here before tracking pen so that the pen thinks they changed at the same time
-    else {
-        this.privateX += steps * Math.cos(this.radians());
-        this.privateY += steps * Math.sin(this.radians());
-      }
+    this.privateX += steps * Math.cos(this.radians());
+    this.privateY += steps * Math.sin(this.radians());
     this.project.ready(this.trackPen);
   };
 
@@ -1068,8 +1062,6 @@ Woof.prototype.Sprite = function () {
   };
 
   this.radians = function () {
-    // subtract 90 from the angle of a line before calculating radians (undoing the correction in the Line() constructor)
-    if (_this.type == "line") return (_this.angle - 90) * Math.PI / 180;
     return _this.angle * Math.PI / 180;
   };
 
@@ -1567,6 +1559,16 @@ Woof.prototype.Line = function () {
       throw new TypeError("You cannot set line.angle directly. You can only modify line.angle by changing the position of the line's points.");
     }
   });
+
+  // using move() with lines is tricky, so throw an error (should probably be fixed at some point)
+  this.move = function () {
+    throw new TypeError("You cannot move lines with move() unfortunately! Change the x, y, x1, and y1 values instead.");
+  };
+
+  // subtract 90 from the angle of a line before calculating radians (undoing the correction in the Line() constructor)
+  this.radians = function () {
+    return (this.angle - 90) * Math.PI / 180;
+  };
 
   this.render = function (context) {
     context.fillStyle = _this7.color;
--- a/dist/woof.js
+++ b/dist/woof.js
@@ -949,8 +949,10 @@ Woof.prototype.Sprite = function () {
     var rotatedY;
     // If sprite is a line, offsets positioning by half the height as line is drawn from endpoints, not center
     if (this.type == 'line') {
-      rotatedX = Math.cos(this.radians()) * (x - this.x) - Math.sin(this.radians()) * (y - this.y + this.height / 2) + this.x;
-      rotatedY = Math.sin(this.radians()) * (x - this.x) + Math.cos(this.radians()) * (y - this.y + this.height / 2) + this.y;
+      // rotatedX = Math.cos(this.radians()) * (x - this.x) - Math.sin(this.radians()) * (y - this.y + this.height / 2) + this.x;
+      // rotatedY =  Math.sin(this.radians()) * (x - this.x) + Math.cos(this.radians()) * (y - this.y + this.height / 2) + this.y;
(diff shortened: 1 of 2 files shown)


stevekrouse:

https://cdn.rawgit.com/stevekrouse/WoofJS/862fc172b4079cf3a2c3424c9be01b8bde25b106/index.html

madduccino:
This looks great - thanks for fixing Joe! I knew that I had switched around the x and y in the atan2 but it seemed to be working so I left it. I'm glad it's working as it should now. 

I actually like thickness!

In regards to length, what purpose does it serve? If the user tries to set the length, an error is thrown and the line is not altered, correct?

madduccino:
Also, if the width property is changed to thickness, that means that all games that have used line width (which is probably not that many) will be affected? eg the tongue for fly game would be 1, the default, as width is no longer recognized

madduccino:
Also just noticed that if I switch the width and height of the rectangle in the test case, the original problem emerges...

`var fourSide = new Rectangle({
  width: 30,
  height: 200
})`

joebeachjoebeach:
Thanks for the feedback, Maddy!

Yeah, I think I wasn't super clear about length vs height etc, so here's a more detailed explanation:

My idea is that length/thickness seem like less ambiguous terms for line properties than height/width, so length/thickness would be the "official" property names -- the ones appearing in the documentation, that is, and therefore the ones we'd "want" users to refer to in their code. Meanwhile, height/width are still there, but only because Woof needs them behind the scenes to render the sprite and deal with collision detection; they're basically "hidden" from the user, but in case a user does try to set them, they return errors.

So, the purpose of `length` in this proposal would be the same as the purpose of `height` in the current version -- the user could get the value of `length` and use it for a variety of things. In the tutorial I'm working on, the length of a line drawn by the mouse affects the power of a projectile, for example.

I suppose, to address the issue of width/thickness conflict with older games, I could make them both do the same thing:
```js
Object.defineProperty(this, 'thickness', {
  get: function() {
    return this.lineThickness;
  },
  set: function(value) {
    if (typeof value != "number") { throw new TypeError("line.thickness can only be set to a number."); }
    this.lineThickness = value;
  }
});

Object.defineProperty(this, 'width', {
  get: function() {
    return this.lineThickness;
  },
  set: function(value) {
    if (typeof value != "number") { throw new TypeError("line.width can only be set to a number."); }
    this.lineThickness = value;
  }
});
```
It's a little clunky to have redundant properties, though, so maybe it's best to keep `width` and ditch `thickness`?

As far as switching the height/width of the rectangle in the test case, I feel like that's ok. If a user wants a rectangle whose short face is "forward", then they need `width` to be longer than `height`, and if they want its long face "forward", then they need `height` to be longer than `width`.

joebeachjoebeach:
Another thing:

In the current commit on this branch, `line.angle` returns the angle **from** (`line.x1`, `line.y1`) **to** (`line.x`, `line.y`), but I feel like it should be the other way around. For some reason, though, I couldn't get it to render properly when I did that. 

What do you guys think? More than happy to work it out, but if you guys think the current version is OK, I don't want to waste time on it.

stevekrouse:
Great conversation! I agree with @joebeachjoebeach  that the line API can be better worded. 

If we could redesign it from scratch, I like `x1`, `y1`, `x2`, `y2`, `thickness`, and `color`. You can read-only `angle` which points from `x1`, `y1` to x2`, `y2` and the length of the line. The `height` of a line returns an error because lines have no height. 

However, as @madduccino says,  given that we don't want to push any breaking changes to Woof, we have to be careful with how we go about this, potentially still accepting `x`, `y`, and `width` parameters so old code still works. My vote would be to focus the scope of this particular pull request to simply getting the angle of lines in line with their endpoints, pointing from `x`, `y` to `x1`, `y1` and leave the API changes for a separate issue / pull request.

(Also, I'm not clear on what  the issue is with `var fourSide = new Rectangle({ width: 30, height: 200 })`...?)

As terribly ugly as it sounds, I'd be happy for a simple workaround where you add 90 degrees to the angle of lines and then correct for this in the line render and collision methods by subtracting the 90 degrees if the `sprite.type == "line"`.

joebeachjoebeach:
Thanks for the input and guidance, @stevekrouse 

I'll follow the "ugly" approach then!

I also just noticed that moving a line with the `move()` method actually rotates it around one end: (http://woofjs.com/create.html#line-move) -- is that desired? If not, maybe we can make a new issue for that.

stevekrouse:
Haha that's actually hilarious -- move changes the x and y positions of a sprite according to it's angle, but a line's angle is 90 degrees off from the line so that makes it move in a circle. Once this issue is solved, move will make the line shorter from it's x, y endpoints towards its x1, y1 endpoints.

However... you're probably right that we don't want to allow this and we can override move in the line sprite with an error. Lines are such trouble!

joebeachjoebeach:
Okey doke -

I added 90 to the angle in the `Line()` constructor, subtracted it when necessary, and made `move()` return an error if `this.type == "line"`.

The collider and renderer both use `rotatedVector(x, y)` to do the rotation, and that uses `radians()` -- so, I subtracted the 90 degrees from the angle within the `radians()` function `if (this.type == "line")`. Is that ok? It looks like we don't really expect users to use `radians()` as it's not in the documentation, and, as of now, every time we use `radians()` for a line sprite in the codebase, we need to remove those 90 degrees. If you'd rather I leave the `radians()` function alone and do this calculation within the `rotatedVector(x, y)` function each time `this.radians()` is called for a line, let me know.

stevekrouse:
I agree with your reluctance to edit the `radians()` function, however it probably does what we want because that's what is used for `rotatedVector()` and for rotating in `render()`.

You can override the `move()` method in the `Line` function itself and not even touch the `Sprite` class. Simply make a `this.move = funciton() { ... }` with the error message and your comment. And the same could go for `radians()` in the Line class. I think this would be cleaner because we wouldn't have to edit the `Sprite` function at all!

joebeachjoebeach:
That makes sense! Done and done.

Issue #318, opened 2017-03-03
woof range() function to replace for-loops entirely


stevekrouse:
For loops are the worst. Wouldn't it be magical if we could do away with them forever?

The solution I have in mind is to create a `range()` function that produces an array that you can forEach over.

It takes 3 params, start, stop (and it stops on stop, not before it, guaranteeing that all numbers in the return array are `>= Math.min(start,stop)` and `<= Math.max(start,stop)`), and step. If stop is more than start, it assumes step is -1, otherwise it assumes step is 1. You can also supply a step/

```javascript
range(1, 5)
=> [1,2,3,4,5]

range(5, 1)
=> [5,4,3,2,1]

range(2, 10, 2)
=> [2,4,6,8,10]
```

This has a few benefits:

1. Not having to learn another syntax when students are already familiar with forEach loops. It also reinforces the forEach syntax.
2. You can easily print out the result of a `range()` as a value so you can see if it does what you inspect. This is a little easier than adding a console log in the body of a for loop.

Lastly, I stole the name "range" from python I think but I'm not in love with it. I'd be super open to other ideas for the name or how the parameters work.

stevekrouse:
There's a large chance this already exists natively in JavaScript or someone has made a library for it already, so it'd make sense to spend some time looking for it.

stevekrouse:
https://lodash.com/docs/4.17.4#range

joebeachjoebeach:
I like this idea, but I want to make sure I'm seeing it the way you are
Does it mean you want to replace this:
```js
for (var i = 0, i < 10, i++) {
  /* code to repeat 10 times */
}
```
with this?:
```js
range(1, 10).forEach(thing => {
  /* code to repeat 10 times */
}
```
If the end goal is repetition of a single block of code, would something like this be more intuitive?:
```js
instantRepeat(10, () => {
  /* code to repeat 10 times */
}
```
I could see how keeping track of the number is useful, by using the `thing` parameter with `forEach()` above, but I guess that could be handled by including a parameter that counts as we go:
```js
instantRepeat(10, i, ()=> {
  /* code to repeat here
  the value of i is set to the repetition number */
}
```

I'm guessing there's a whole class of use-cases that I'm not thinking of, as your example of creating the array `[2, 4, 6, 8, 10]` would clearly be used for something other (more?) than repeating code 5 times.

[...]

PR #406, opened 2017-06-14
create makeArray function

joebeachjoebeach:
addresses #318 

I named the function `makeArray()` -- how does that sound?


stevekrouse:

https://cdn.rawgit.com/stevekrouse/WoofJS/06635c75c00ef7108d5b6a2fb68ad30b20f7b81e/index.html

src/woof.es6.js
```diff
@@ -1477,6 +1477,24 @@ Woof.prototype.mobile = function() {
     return (typeof window.orientation !== "undefined") || (navigator.userAgent.indexOf('IEMobile') !== -1);
 };
 
+// generate an array with the given parameters (starting value, ending value, incrementation) default incrementation is 1
+Woof.prototype.makeArray = function(start, end, incr = 1) {
+  var output = [];
+  if (start < end) {
+    incr = Math.abs(incr);
+    for (var i = start; i < end + 1; i += incr) {
```
stevekrouse: I think the terminating condition we want here is `i <= end`, no?

src/woof.es6.js
```diff
@@ -1477,6 +1477,24 @@ Woof.prototype.mobile = function() {
...
+      output.push(i);
+    }
+  }
+  else if (start > end) {
+    incr = -Math.abs(incr);
+    for (var j = start; j > end - 1; j += incr) {
```
stevekrouse: Same here, `j >= end`

src/woof.es6.js
```diff
@@ -1477,6 +1477,24 @@ Woof.prototype.mobile = function() {
     return (typeof window.orientation !== "undefined") || (navigator.userAgent.indexOf('IEMobile') !== -1);
 };
 
+// generate an array with the given parameters (starting value, ending value, incrementation) default incrementation is 1
+Woof.prototype.makeArray = function(start, end, incr = 1) {
```
stevekrouse: I'm not in love with `makeArray`. In particular, the word "make" is not very JavaScript-y. You'd normally say `new Array()` or something like that.

I sorta like `numbersFrom(1, "to", 10, "by", 1)` like we do with `every`s but that's *really* not JavaScript-y.

While I do like that including the word array makes it explicit that this is creating an array, it doesn't explain that it's creating a range of numbers which is important.

While the word "range" isn't perfect for kids, it's the most descriptive I can think of still. I also don't hate the fact that kids will be familiar with it if they go to python, although that's not reason enough to do it on it's own.

Another option could be `numbers(1, 10)`. Or `series(1, 10)`. Or `sequence(1, 10)`. Thoughts? Other ideas?

joebeachjoebeach:
I kept the name `range()` because I think it's the best fit. I also made the requested edits. Let me know your thoughts!

stevekrouse:
Looks great! I agree that range is a good name.

Can you add to the docs?

joebeachjoebeach:
I made a simple entry in the docs, but I'm not sure if it's enough. Do we need examples? Part of me wants to show what the resulting array would be. For example:

Create an array with a range of numbers from one to ten - [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]:
```javascript 
var oneToTen = range(1, 10, 1)
```

Create an array with a range of numbers from 30 to 42 by 3 - [30, 33, 36, 39, 42]:
```javascript 
var thirtyToFortyTwo = range(30, 42, 3)
```

stevekrouse:
Yeah, I like the idea of showing the result, but not sure where to put it in a place that doesn't get in the user's way. Have any ideas?

Another thought is that you can show them how to use this in the context of a `.forEach()` loop. You can even mention explicitly that we recommend `range(1, 10).forEach(i => {})` to the equivalent for-loop.

joebeachjoebeach:
Yeah it's tricky to figure out where to put this. In the context of creating an array with a range of numbers, the "Data" section seems to make the most sense. But, in the context of a replacement for a for-loop, "Control" seems to make more sense.

I also feel like this is a topic that may need a little more explaining rather than the main formula we have in the docs of *Action*: **Code Example**

I would lean towards something like this:

We add this to the "Data" section:

> You can automatically generate an array with a range of numbers by using the `range()` function
>
> Example 1: Generate this array: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
> It starts at 1, ends at 10, and goes up by 1 every time:
> ```javascript
> range(1, 10, 1)
>```
>
>Example 2: Generate this array: [15, 10, 5, 0]
> It starts at 15, ends at 0, and goes down by 5 every time:
>```javascript
>range(15, 0, 5)
>```
>

And we add this to the "Control" section:
> When you use `repeat()`, your code doesn't repeat as fast as the computer can run it. Instead, we slow it down, which makes things like animation and motion work well. But, if you want your code to repeat instantly, as fast as the computer can do it, you can combine the `range()` function with the `forEach()` function:
>
>Example: Repeat your code instantly 10 times:
>```javascript
>range(1, 10).forEach(i => { /* code to repeat */ })
>```
>In the above example, `i` will start at 1 and go up by 1 every repetition until it gets to 10.
>
>If you are familiar with `for`-loops, you might realize that this does the same thing. We recommend using `range().forEach()` instead of `for`.

stevekrouse:
I love these suggestions! Really well written.

One question. `range(15, 0, 5)` feels a little strange. `range(15, 0, -5)` seems to make more intuitive sense to me. Would it make sense to switch it?

ekenberg:
> range(15, 0, 5) feels a little strange. range(15, 0, -5) seems to make more intuitive sense to me.

Yes indeed, it makes sense to use the [same syntax as Python](http://pythoncentral.io/pythons-range-function-explained/)

stevekrouse:
That's a good point @ekenberg! Makes me wonder if we should mirror how Python does strictly less than the upper bound while our range function does less than or equal to. Personally I prefer less than or equal to, but I could see the argument for consistency with an existing function.

joebeachjoebeach:
@stevekrouse @ekenberg totally agree about the intuitiveness of `range(15, 0 -5)` -- should we throw an error, then, if someone tries to do something like `range(15, 0, 5)`?

stevekrouse:
I think we can figure out what they mean for `range(15, 0, 5)` and not throw the error. However, in the example, we can use the proper `range(15, 0, -5)` syntax

ekenberg:
@stevekrouse - this is a potentially [hot topic](https://stackoverflow.com/questions/11364533/why-are-slice-and-range-upper-bound-exclusive). At first glance, the half-open interval can feel non-intuitive, but it has its merits. And who wants to argue with [Dijkstra...](http://www.cs.utexas.edu/users/EWD/ewd08xx/EWD831.PDF)

I guess as soon as you start inventing or adding new constructs to a language, there is a lot of history and science to relate to. There's the balance between being "intuitive" and scientifically correct. If Woof range() would include the upper bound for "intuition's" sake - should you also start indexing arrays from 1 and not 0? _There be dragons..._

ekenberg:
@joebeachjoebeach @stevekrouse - for `range(15, 0, 5)` python will simply do nothing, which seems to me the correct behaviour. When the parameters to `range(a,b,c)` are variables, messing with the step-sign can introduce all kinds of weird bugs or unwanted behaviour. Predictability and consistency should be more important than being overly helpful.

ekenberg:
Maybe, if you want to be very instructive, have the editor show a note/warning when a user uses static arguments to range as in `range(100, 0, 1)` _<-- Some friendly message about using a negative step_

stevekrouse:
Very good points. I could truly be convinced either way.

In the longer term, I'd love to have 1-indexed arrays and fully closed intervals, because while they might not be as mathematically elegant, they are much clearer to new programmers.

However, in the short term, I do see Woof as a transition tool to "real programming" so I guess we might as well be consistent with it when it's not too inconvenient.

I'll leave it up to you, @joebeachjoebeach to make the final call.

Regarding not doing something when the parameters are put in the wrong order... I'd love to either do what we expect they want (which in this particular case seems easy to detect without much unwanted behavior) or throw an error like Joe suggested or add an in-editor warning like @ekenberg suggested. Because I'd hate for a kid to have it returning an empty array and be totally confused, especially when we're 99% certain what her mistake was.

joebeachjoebeach:
OK. So, I made the second parameter exclusive, as I think it's best to be consistent with 0-indexing etc.

I was going to opt to throw an error in a case like `range(10, 0, 1)`, but I realized it would break the functionality of the two-parameter option. That's because `range(0, 10)` would work, but `range(10, 0)` would throw an error. Since the two-parameter option was part of @stevekrouse's original suggestion in #318, and because I like that functionality, I'd like to preserve it. It would be messy to assume they wanted to use -1 in that case, but not assume that they wanted to use -1 in the case of `range(10, 0, 1)`. So, I opted to have Woof make the educated guess instead of throwing an error, allowing `range(10, 0, 1)` and `range(10, 0, -1)` to output the same thing.

stevekrouse:
💥 👊 🎆 🎉

ekenberg:
Nice work @joebeachjoebeach 
However, I couldn't stop thinking about this. Arguing that Dijkstra and Python matters felt good at the time but I didn't convince myself :)

If I understand correctly, `range(1,5)` will now yield `1,2,3,4`? I can't even begin to think how to explain that to a child just emerging from Scratch: _Ok so you have five fingers, right? To count them in WoofJS you must use a `foreach` on the `range` function and go from `1 to 6`..._

Then I came to think of list generation in Haskell:
`[1..5] -- yields 1,2,3,4,5`
`[1,3..9] -- yields 1,3,5,7,9`
`[5,4..1] -- yields 5,4,3,2,1`

Notice how it's inclusive at both ends, and how it intuitively picks up the step-size when it differs from 1. `[1..5] -- five fingers` makes sense.

There's even a [range function](http://hackage.haskell.org/package/base-4.10.0.0/docs/Data-Ix.html#v:range) in Haskell. It works like this:
`range (1,3) -- yields 1,2,3`: yes inclusive at both ends too.

Haskell is a [favorite among computer scientists](https://www.quora.com/Why-do-computer-scientists-like-Haskell-so-much). If Haskell can be simple and intuitive, WoofJS can too!

At this point, I'm going to argue that WoofJS `range` should be inclusive at both ends, and as clever as possible with regards to step-sizes. Be helpful to the kids. When they outgrow Woof and turn to Python, they'll be mature enough to handle the syntactic change. But if Woof doesn't make intuitive sense to them, they might drop programming all together.

@stevekrouse how do you feel about this, from the perspective of teaching kids?

stevekrouse:
Thanks for continuing to think on this @ekenberg! I can really see the argument from either perspective. We just have to pick one and stick with it, so if we want to make this change, we should do it ASAP before students start using the range function.

If you'd like to change it, please shoot over a pull request :)

stevekrouse:
Also relevant to this conversation... http://alarmingdevelopment.org/?p=470

Issue #91, opened 2016-05-25
add Ellipse w height and width


stevekrouse:

stevekrouse:
Or rectangle with rounded edges?

stevekrouse:
Circles are too limited of a sprite. We should probably change them to ellipses. My comment above ^ was wondering if ellipses and rectangles are so similar (width and height) that they should be the same sprite. Probably not but something worth thinking about.

Issue #268, opened 2016-12-05
more accurate colliders


stevekrouse:

stevekrouse:
Given your recent experience with collides, @madduccino, I think this could be a good issue for you.

https://github.com/jriecken/sat-js#satcircle

stevekrouse:
Circles should have circle colliders. https://github.com/jriecken/sat-js#satcircle
 
Rectangles should have rectangle colliders.

Polygons (which are coming soon) should have polygon colliders.

PR #410, opened 2017-06-15
added RegularPolygon and Oval sprites

jwass91:
The autocomplete for these sprites didn't seem to work. 
Check out the vendor/codemirror/javascript-hint.js


stevekrouse:

https://cdn.rawgit.com/stevekrouse/WoofJS/0fd60c5ee71af041ff2ed297a5d035a2125cbf98/index.html

stevekrouse:
Will close #91 and #131

src/woof.es6.js
```diff
@@ -1062,6 +1062,89 @@ Woof.prototype.Rectangle = function({project = undefined, height = 10, width = 1
...
+  
+  this.render = (context) => {
+    context.fillStyle=this.color;
+    context.beginPath();
+    context.ellipse(0, 0, width/2, height/2, 0, 0,  2*Math.PI);
+    context.fill();
+  };
+};
+
+Woof.prototype.RegularPolygon = function({project = undefined, sides = 3, length = 100, color = "black"} = {}) {
```
stevekrouse: Maybe let's call this `Polygon`? I don't think we'll be adding irregular polygons to Woof anytime soon, and if we do we can call them `IrregularPolygon`s

src/woof.es6.js
```diff
@@ -1062,6 +1062,89 @@ Woof.prototype.Rectangle = function({project = undefined, height = 10, width = 1
...
+    get: function() {
+      return this.regularpolygonLength;
+    },
+    set: function(value) {
+      if (typeof value != "number") { throw new TypeError("regularpolygon.length can only be set to a number."); }
+      this.regularpolygonLength = value;
+    }
+  });  
+
+  this.collider = function() {
```
stevekrouse: Why are you overriding `collider` but not actually changing it? That is, you can delete this method here and it will still work the same way, yes?

(We're going to override this collider eventually in #268 but I thought we decided not worry about that on this pull request.)

docs/index.html
```diff
@@ -458,7 +458,37 @@
                         }, {
                             description: "Set color:",
                             code: "color: rgb(100, 50, 240)"
-                        }, ]
+                        } ],
+                         RegularPolygon: [{
```
stevekrouse: Looking at the order of these menus, I think I'd like to switch it up a bit. Can you make the order:

Background
Text
Image
Rectangle
Polygon
Circle
Oval
Line

stevekrouse: I think it'd also be good for all of these to be on the screen at the same time without having to scroll so maybe make the height of these buttons a tad smaller?

PR #431, opened 2017-07-08
New Login/Signup

jwass91:
Added new login/signup modal that has:
- unique username and email requirements
- signup with option to 'receive updates about WoofJS'
- reset password capabilities
- helpful error messages

Issue: It currently first validations username before going onto to validate emails so the email error will not show up until a valid username is entered.


stevekrouse:

https://cdn.rawgit.com/stevekrouse/WoofJS/d8395ada9ca3a9b375d9aaeeb00cfb372a5c5284/index.html

stevekrouse:
* when I type an illegal character in my username, the input box and message are green which seems off. Maybe just have them have no color?
* when I type any character into the email address field, it shakes and gives me a message to type a valid email address. It seems like this field should be a little more patient. Potentially, it shouldn't react until you try to submit it and the email address is still invalid? 
* ditto for the password field. It could be more patient. Potentially it could give those errors on blur, when we click out of that input box?

stevekrouse:
Make sure to delete the following files that will no longer be used:

- [ ] https://github.com/stevekrouse/WoofJS/blob/master/login.html
- [ ] https://github.com/stevekrouse/WoofJS/blob/master/vendor/firebase/firebase-ui-auth.js (you can delete the folder containing this file as well)

create.html
```diff
@@ -230,8 +228,13 @@
 </head>
 <body>
   <div id="page">
-    <div id="navbar"> 
-      <button style="background-color: #80CFFF; color: white;  border-radius:4px; border:none; margin:8px" class="btn btn-medium btn-info" type="button" id="login" onclick="window.open('./login.html', 'Sign In', 'width=500,height=500')">Login / Sign Up</button>
+    <div id="navbar">
+
+    <div class="btn-group" role="group" aria-label="..." id='loginSingupGroup'>
```
stevekrouse: I'm not crazy about how these buttons look...

* maybe they would look better not as a button group but as two separate buttons?
* I don't love the color of the sign-up button. Logically, `btn-warning` doesn't really make sense. Maybe `btn-success`? Maybe `btn-info`? Maybe have them be the same color? Play around with it :)

create.html
```diff
@@ -273,8 +276,119 @@ <h4 class="modal-title" id="saveModalLabel">Choose a name for your project</h4>
...
+              <label for='username'>Email</label>
+              <input type="text" v-model="loginEmail" class="form-control" id="username" placeholder="Email" aria-describedby="usernameHelp">
+              <span id="usernameHelp" class="help-block" :style="{display: loginError == 'USERNOTFOUND' ? 'block' : 'none'}">No account with this email exists. Please try to sign up for an account.</span>
+               <span id="usernameHelp" class="help-block" :style="{display: loginError == 'INVALIDEMAIL' ? 'block' : 'none'}">Invalid email address. Please try again.</span>
+            </div>
+            <div class="form-group" :class="{'has-error': loginError == 'WRONGPASSWORD',  animated: loginError == 'WRONGPASSWORD',  shake: loginError == 'WRONGPASSWORD'}">
+              <label for='password'>Password</label>
+              <input onkeydown="if (event.keyCode == 13){login()}"  type="password" v-model="loginPassword" class="form-control" id="password" placeholder="Password" aria-describedby="passwordHelp">
```
stevekrouse: You can do the onkeydown like this: `@keydown.enter="login()"`

create.html
```diff
@@ -273,8 +276,119 @@ <h4 class="modal-title" id="saveModalLabel">Choose a name for your project</h4>
...
+            </div>
+            <div class="form-group" :class="{'has-success': passwordValid == 'VALID', 'has-error': passwordValid == 'WEAK','animated': passwordValid == 'WEAK','shake': passwordValid == 'WEAK'}">
+              <label for='username'>Password</label>
+              <input type="password" v-model='signUpPassword' class="form-control" id="passwordCreate" placeholder="Password" aria-describedby="passwordCreateHelp" @input="passwordValidate($event)" onkeydown="if (event.keyCode == 13){signUp()}">
+              <span id="passwordCreateHelp" class="help-block" :style="{display: passwordValid == 'WEAK' ? 'block' : 'none'}">Passwords must have at least 6 characters and a mix of letters and numbers.</span>
+             <!--  <input  onfocus="$('#passwordCreate').tooltip('toggle')" data-placement="top" title="Remember it" type="password" class="form-control" id="passwordCreate" placeholder="Password" aria-describedby="passwordCreateHelp" @input="passwordValidate($event)" onkeydown="if (event.keyCode == 13){signUp()}"> -->
```
stevekrouse: Can you delete this or are you saving it here for a reason?

create.html
```diff
@@ -387,15 +501,35 @@ <h4 class="modal-title" id="saveModalLabel">Choose a name for your project</h4>
         mouseX: 0,
         mouseY: 0,
         seenError: true,
-        
+
         projectNameRequest: "NOT LOADING", // or "LOADING" or "NAME TAKEN"
         projectNameInputValue: "",
         projectNameLastKeyIllegal: false,
+        usernameLastKeyIllegal: false,
+        loginSignUpView: "LOGIN", //OR SIGNUP OR RESETPASSWORD
```
stevekrouse: Could you add a space after all of these `//`? Also could you lowercase the "OR"s on this line like you do on most of the other lines? You can lowercase the "OR"s on a few other lines too.

create.html
```diff
@@ -428,32 +562,82 @@ <h4 class="modal-title" id="saveModalLabel">Choose a name for your project</h4>
...
+          this.loginSignUpView = 'LOGIN'
+        },
+        switchToPasswordReset: function(){
+           this.loginSignUpView = 'RESETPASSWORD'
+        },
+        usernameValidate: function($event){
+          var username = $event.target.value
+          if (username == ''){
+            this.usernameValid = "EMPTY"
+          }else{
```
stevekrouse: Add a space on either side of this else, like this: `} else {`. You can do this with other else's in your code on lines  624, 626, 634, 636, etc.

create.html
```diff
@@ -428,32 +562,82 @@ <h4 class="modal-title" id="saveModalLabel">Choose a name for your project</h4>
...
+            this.emailValid = "NOTEMAIL"
+          }else if(/^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/.test(email)){
+            this.emailValid = "VALID"
+          }
+        },
+        passwordValidate: function($event){
+          var password = $event.target.value
+          if (password == ''){
+            this.passwordValid = "EMPTY"
+          }else if(!/(?=.*[a-zA-Z])(?=.*\d)(?=.{6}).*/.test(password)){
```
stevekrouse: Good thinking! However, most of what you think you know about password rules are bullshit:

https://blog.codinghorror.com/password-rules-are-bullshit/

However, given that we're not fort knox here, we don't have to be so careful with things. My bias is to require the password be 6 characters long and that's it -- no mix of numbers and letters. If we really wanted to be secure, we could increase the length limit to 10 characters but I think that's unnecessary. What do you think?

create.html
```diff
@@ -475,42 +659,157 @@ <h4 class="modal-title" id="saveModalLabel">Choose a name for your project</h4>
...
+      app.loginPassword = "";
+      app.loginEmail = "";
+      app.resetEmail = "";
+      app.signUpEmail = '';
+      app.signUpPassword = '';
+      app.signUpChecked = true;
+      app.signUpUsername = '';
+    }
+
+    function login() {
```
stevekrouse: Looks like a really well-written function!

create.html
```diff
@@ -475,42 +659,157 @@ <h4 class="modal-title" id="saveModalLabel">Choose a name for your project</h4>
...
+        app.resetButtonState = "NORMAL"
+      })
+    }
+
     function handleLogin(userName, uid) {
       $('#loggedin-name').html(userName + ' <span class="caret"></span>')
       $('#login-projects').attr("href", "./user.html#" + uid )
       $('#logggedin').removeClass("hide")
       $('#login').addClass('hide')
+      $('#signup').addClass('hide')
```
stevekrouse: This entire function `handleLogin` and its sister `handleLogout` are very un-VueJS. In theory, it'd be great to delete them entirely and put all of this logic into the HTML/VueJS so that the buttons "know themselves" whether they should be showing or hiding. Clearly, I didn't think it was important or easy enough to do myself but you can feel free to give it a shot if you want to.

create.html
```diff
@@ -475,42 +659,157 @@ <h4 class="modal-title" id="saveModalLabel">Choose a name for your project</h4>
...
-          if (!snapshot.val()) {
-            firebase.database().ref("/user/" + user.uid).set({
-              displayName: user.displayName,
-              email: user.email,
-              createdDate: firebase.database.ServerValue.TIMESTAMP
-            })
-          }
-        })
-        
+        if(user.displayName){
```
stevekrouse: space after if

create.html
```diff
@@ -475,42 +659,157 @@ <h4 class="modal-title" id="saveModalLabel">Choose a name for your project</h4>
...
+          handleLogin(user.displayName, user.uid)
+        }else if(app.signUpUsername){
```
stevekrouse: space before else and after if

create.html
```diff
@@ -475,42 +659,157 @@ <h4 class="modal-title" id="saveModalLabel">Choose a name for your project</h4>
...
+
+                var user = firebase.auth().currentUser;
+
+                user.updateProfile({
+                  displayName: app.signUpUsername
+                })
+
+                $('#loggedin-name').html(app.signUpUsername + ' <span class="caret"></span>');
+
+                firebase.database().ref("uniqueUsername/" + app.signUpUsername).set({
```
stevekrouse: Nicely done!

create.html
```diff
@@ -475,42 +659,157 @@ <h4 class="modal-title" id="saveModalLabel">Choose a name for your project</h4>
...
+                  username: app.signUpUsername,
+                  uid: user.uid
+                })
+
+                firebase.database().ref('/user/' + user.uid).once('value').then(function(snapshot) {
```
stevekrouse: Perfect 💯

create.html
```diff
@@ -475,42 +659,157 @@ <h4 class="modal-title" id="saveModalLabel">Choose a name for your project</h4>
...
+
+                $('#loginModal').modal('hide');
+
+                var user = firebase.auth().currentUser;
+
+                user.updateProfile({
+                  displayName: app.signUpUsername
+                })
+
+                $('#loggedin-name').html(app.signUpUsername + ' <span class="caret"></span>');
```
stevekrouse: In theory, this is not very VueJS-like. Whenever you use the JQuery `$`, in theory we could do it with Vue "better". However, as long as this works it's not the end of the world, so no need to change it if you don't want to.

create.html
jwass91: I can try making them two buttons but the reason I went with a button group is because I was having trouble getting them to align nicely when they were two individual buttons. Do you like this look?
screen shot 2017-07-24 at 2 02 21 pm

create.html
jwass91: The `login()` function is a javascript function not a vue function so `@keydown.enter="login()"` doesn't work. I have it as a javascript function because the firebase stuff doesn't work in a vue function.

create.html
jwass91: The reason I added this is because firebase requires the passwords to fit this criteria. Otherwise, firebase will throw an error when I go to sign up the user. In the current sign up for woof, the same password rules apply. Maybe there is a way to change the firebase settings for this?

create.html
stevekrouse: Yeah, that looks great!

create.html
stevekrouse: Ok, sounds good!

create.html
stevekrouse: It seems like they only check if it's less than 6 chars. Is that not the case?

https://firebase.google.com/docs/reference/android/com/google/firebase/auth/FirebaseAuthWeakPasswordException

create.html
```diff
@@ -475,42 +675,165 @@ <h4 class="modal-title" id="saveModalLabel">Choose a name for your project</h4>
...
+      } else {
+        app.passwordValid = "VALID"
+      }
+
+      if (app.usernameValid == 'VALID' && app.emailValid == 'VALID' && app.passwordValid == 'VALID' && app.signUpButtonState != 'LOADING') {
+        app.signUpButtonState = "LOADING"
+
+        var exists
+        var ref = firebase.database().ref("uniqueUsername");
+        ref.child(app.signUpUsername).once('value', function(snapshot) {
```
stevekrouse: I'm realizing now that we have a potential issue here... There's a small but non-zero chance that between when you check if the unique username exists in the database and when you store the unique username in the database, another user is creating that exact same username and beats you to it.

The solution here is that instead of querying the uniqueusername field, we try to update it first. If it succeeds, we're good and if it fails, we know that someone already has that username. (It would fail if someone already has that username because that's how the rules we set up in firebase work, I think.)

stevekrouse:
Looks like you have some debugging to do. When I tried to sign up, it wouldn't let me, giving me a "Passwords must be at least 6 characters long." error no matter if my password was over 6 characters.

create.html
stevekrouse: You're welcome to create a separate issue for this and do it after this pull request merges. It's important but not *that* important.

create.html
```diff
@@ -475,42 +675,165 @@ <h4 class="modal-title" id="saveModalLabel">Choose a name for your project</h4>
...
+      })
+    }
+
     function handleLogin(userName, uid) {
-      $('#loggedin-name').html(userName + ' <span class="caret"></span>')
-      $('#login-projects').attr("href", "./user.html#" + uid )
-      $('#logggedin').removeClass("hide")
-      $('#login').addClass('hide')
+      app.userNameForMenu = userName
+      app.uid = "./user.html#" + uid
```
stevekrouse: This variable `app.uid` isn't a great name. I'd prefer you called it `app.userProjectsLink to be more accurate`.

Or if you want to leave it as `app.uid = uid`, that works too, so then you'd have to change it above to `:href="'./user.html#' + uid"`

create.html
jwass91: I'll take care of this in another issue. If I remember correctly, I had tried doing this originally and the issue was that it just would override the username if it existed rather than throwing an error.

create.html
```diff
@@ -428,32 +566,94 @@ <h4 class="modal-title" id="saveModalLabel">Choose a name for your project</h4>
...
+          var password = $event.target.value
+          if (password == ''){
+            this.passwordValid = "EMPTY"
+          } else if (!/^.{6,}$/.test(password)){
+            this.passwordValid = "WEAK"
+          } else {
+            this.passwordValid = "VALID"
+          }
+        },
+        emailSuccessValidate: function($event){
```
stevekrouse: What's the difference between `emailValidate` and `emailSuccessValidate`?  

Can you add a comment or two explaining the difference?

create.html
```diff
@@ -428,32 +566,94 @@ <h4 class="modal-title" id="saveModalLabel">Choose a name for your project</h4>
...
+          var email = $event.target.value
+          if (/^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/.test(email)){
+            this.emailValid = "VALID"
+          }
+        },
+        passwordSuccessValidate: function($event){
```
stevekrouse: Same here... why a seperate success method?

create.html
stevekrouse: Great! https://github.com/stevekrouse/WoofJS/issues/451

create.html
jwass91: The general validate method throws errors and success on the blur event. The success method is there to validate on every input. This way there aren't constant errors but it will still give positive validation at any point in time.

PR #452, opened 2017-07-25
firebase saves user theme

nicolekelner:


create.html
```diff
@@ -326,10 +349,29 @@ <h4 class="modal-title" id="saveModalLabel">Choose a name for your project</h4>
       </div>
       <div v-show="codePane" :style="{display: !codebar ? 'flex' : 'none'}" id="code">
         <div id="CodeNav">
+          <div>
           <span style="margin-left:8px; margin-top:4px">Code</span>
+               <select v-model="selected"  style="color:#49A6DF; margin-top: 4px; height: 18px; font-size: 12px;">
+                 <option disabled value="" id="select">Select Theme</option>
+              		<option>monokai</option>
```
stevekrouse: these options need to be unindented

create.html
```diff
@@ -326,10 +349,29 @@ <h4 class="modal-title" id="saveModalLabel">Choose a name for your project</h4>
...
+              		<option>erlang-dark</option>
+              		<option>lesser-dark</option>
+              		<option>eclipse</option>
+              		<option>3024-night</option>
+              		<option>base16-dark</option>
+              		<option>elegant</option>
+              		<option>midnight</option>
+              		<option>night</option>
+              	</select>
+              	</div>
```
stevekrouse: format of these closing tags seems off

create.html
```diff
@@ -387,7 +429,7 @@ <h4 class="modal-title" id="saveModalLabel">Choose a name for your project</h4>
         mouseX: 0,
         mouseY: 0,
         seenError: true,
-        
+        selected: '',
```
stevekrouse: this variable name needs to be more descriptive

create.html
```diff
@@ -454,13 +496,25 @@ <h4 class="modal-title" id="saveModalLabel">Choose a name for your project</h4>
             }
           }
         },
+        
+       
+        
+        switchTheme : function(){
+          var selected = this.selected;
+          editor.setOption("theme", selected);
+            firebase.database().ref("/user/" + firebase.auth().currentUser.uid +"/theme").set(selected)
```
stevekrouse: this line needs to be unindented

create.html
```diff
@@ -491,12 +545,21 @@ <h4 class="modal-title" id="saveModalLabel">Choose a name for your project</h4>
...
     var authChanged = false
     firebase.auth().onAuthStateChanged(function(user) {
-      if (!authChanged){
-        authChanged = true
-        if(user){
-          ga('set', 'userId', user.uid); 
-        } 
-      }
+
+          //update user theme when new page loads
```
stevekrouse: insert space after //

stevekrouse: Also this comment isn't quite accurate. Please fix.

create.html
```diff
@@ -661,12 +727,16 @@ <h4 class="modal-title" id="saveModalLabel">Choose a name for your project</h4>
...
+    
+    
+    
     var editor = CodeMirror(document.getElementById('code'), {
         mode:  "javascript",
         value: editorVal,
         lineNumbers: true,
         lineWrapping: false,
-        theme: "eclipse",
+        // theme: "eclipse",
```
stevekrouse: either delete this line entirely or uncomment it

create.html
```diff
@@ -491,12 +545,21 @@ <h4 class="modal-title" id="saveModalLabel">Choose a name for your project</h4>
...
-        if(user){
-          ga('set', 'userId', user.uid); 
-        } 
-      }
+
+          //update user theme when new page loads
+          firebase.database().ref('/user/' + firebase.auth().currentUser.uid + "/theme").once('value').then(function(snapshot) {
+             editor.setOption("theme", snapshot.val());
+            var sel = document.getElementById('select');
+            sel.innerHTML = '<option>'+ snapshot.val() +'</option>';
```
stevekrouse: This is not the VueJS way to do this. Use the v-modal. It's easier than you'd think.

Whenever you use the Jquery $ or document.getElementById you're probably not using Vue as much as you could be.

create.html
```diff
@@ -491,12 +545,21 @@ <h4 class="modal-title" id="saveModalLabel">Choose a name for your project</h4>
...
     firebase.auth().onAuthStateChanged(function(user) {
-      if (!authChanged){
-        authChanged = true
-        if(user){
-          ga('set', 'userId', user.uid); 
-        } 
-      }
+
+          //update user theme when new page loads
+          firebase.database().ref('/user/' + firebase.auth().currentUser.uid + "/theme").once('value').then(function(snapshot) {
```
stevekrouse: You probably only want to do this if the user is logged in. `firebase.auth().onAuthStateChanged` runs when the user logs in or out. So you have to check `if (user)` to see if the user is being logged in.

create.html
stevekrouse: More importantly, you only want this code to run if the user is logged in. (If you try your code when not logged in, you'll see that there is an error.)

create.html
stevekrouse: Also, you could look for a place in the code where you already get this value so you don't have to make this redundant firebase call... Hint: it's within 50 lines of code.

create.html
```diff
@@ -577,6 +559,17 @@ <h4 class="modal-title" id="saveModalLabel">Choose a name for your project</h4>
           
         })
         
+         // if the user is logged in, load the code theme they last saved
+          firebase.database().ref('/user/' + firebase.auth().currentUser.uid + "/theme").once('value').then(function(snapshot) {
+             var selectedTheme = snapshot.val() ;
+             editor.setOption("theme", snapshot.val());
+             this.selectedTheme = this.selectedTheme;
+            
+            // var sel = document.getElementById('select');
```
stevekrouse: please delete this comment

create.html
```diff
@@ -58,28 +58,20 @@
...
-<link rel="stylesheet" href="https://codemirror.net/theme/lesser-dark.css">
-<link rel="stylesheet" href="https://codemirror.net/theme/midnight.css">
-<link rel="stylesheet" href="https://codemirror.net/theme/monokai.css">
-<link rel="stylesheet" href="https://codemirror.net/theme/neat.css">
-
-<!--end -->
-  
-  
-  
+  <link rel="stylesheet" href="https://codemirror.net/theme/3024-day.css">
```
stevekrouse: Great!

create.html
```diff
@@ -351,23 +342,23 @@ <h4 class="modal-title" id="saveModalLabel">Choose a name for your project</h4>
...
-              		<option>erlang-dark</option>
-              		<option>lesser-dark</option>
-              		<option>eclipse</option>
-              		<option>3024-night</option>
-              		<option>base16-dark</option>
-              		<option>elegant</option>
-              		<option>midnight</option>
-              		<option>night</option>
+               <select v-model="selectedTheme"  style="color:#49A6DF; margin-top: 4px; height: 18px; font-size: 12px;">
+                <option  placeholder="Select Theme"> {{ selectedTheme }} </option>
```
stevekrouse: The formatting here still needs work

create.html
```diff
@@ -497,23 +488,23 @@ <h4 class="modal-title" id="saveModalLabel">Choose a name for your project</h4>
           }
         },
         
-       
-        
-        switchTheme : function(){
-          var selected = this.selected;
-          editor.setOption("theme", selected);
-            firebase.database().ref("/user/" + firebase.auth().currentUser.uid +"/theme").set(selected)
+        tempTheme: function() {
```
stevekrouse: Why'd you change the name of this function? I liked the last name better, thought it was clearer

create.html
```diff
@@ -497,23 +488,23 @@ <h4 class="modal-title" id="saveModalLabel">Choose a name for your project</h4>
...
+          var selectedTheme = this.selectedTheme;
+          editor.setOption("theme", this.selectedTheme);
+          if (firebase.auth().onAuthStateChanged) {
```
stevekrouse: I don't think checking if the `firebase.auth().onAuthStateChanged` function exists does what you expect here.

create.html
stevekrouse: Also, you don't need a separate function here. You can put all of the code here in the `selectedTheme` watch method.

create.html
```diff
@@ -577,6 +559,17 @@ <h4 class="modal-title" id="saveModalLabel">Choose a name for your project</h4>
           
         })
         
+         // if the user is logged in, load the code theme they last saved
```
stevekrouse: 1. This code doesn't check if the user is logged in like this comment promises

2. There is a place in the code where we are already pulling the theme data (along with other user data), so we don't need to make this extra firebase call. It's right above this section of code.

create.html
```diff
@@ -491,12 +536,12 @@ <h4 class="modal-title" id="saveModalLabel">Choose a name for your project</h4>
     
     var authChanged = false
     firebase.auth().onAuthStateChanged(function(user) {
-      if (!authChanged){
-        authChanged = true
-        if(user){
-          ga('set', 'userId', user.uid); 
-        } 
-      }
+          if (!authChanged) {
```
stevekrouse: you messed up the formatting of this section. Please put it back the way it was

create.html
```diff
@@ -506,11 +551,25 @@ <h4 class="modal-title" id="saveModalLabel">Choose a name for your project</h4>
             firebase.database().ref("/user/" + user.uid).set({
               displayName: user.displayName,
               email: user.email,
-              createdDate: firebase.database.ServerValue.TIMESTAMP
+              createdDate: firebase.database.ServerValue.TIMESTAMP,
```
stevekrouse: please put this back the way it was

create.html
```diff
@@ -506,11 +551,25 @@ <h4 class="modal-title" id="saveModalLabel">Choose a name for your project</h4>
...
+            
             })
           }
+
```
stevekrouse: please put this back the way it was

create.html
stevekrouse: and the one below it

create.html
```diff
@@ -661,12 +720,15 @@ <h4 class="modal-title" id="saveModalLabel">Choose a name for your project</h4>
       cm.replaceSelection(spaces, "end", "+input");
     }
     
+
```
stevekrouse: delete these new lines

create.html
```diff
@@ -696,7 +758,8 @@ <h4 class="modal-title" id="saveModalLabel">Choose a name for your project</h4>
         save();
       }
     }
-    
+
```
stevekrouse: delete these newlines

create.html
```diff
@@ -326,10 +340,29 @@ <h4 class="modal-title" id="saveModalLabel">Choose a name for your project</h4>
...
         <div id="CodeNav">
-          <span style="margin-left:8px; margin-top:4px">Code</span>
-          <span style="margin-right:8px">
-            <img @click="codebar = !codebar" style="margin: 5px; width: 16px; height:18px; cursor:pointer" src="./images/minimize.png" title="Minimize">
-          </span>
+           <div>
+            <span style="margin-left:8px; margin-top:4px">Code</span>
+               <select v-model="selectedTheme"  style="color:#49A6DF; margin-top: 4px; height: 18px; font-size: 12px;">
+                  <option  placeholder="Select Theme"> {{ selectedTheme }} </option>
+                	<option>monokai</option>
```
stevekrouse: you need to un-indent all of these options so they align with the Select Theme option

create.html
```diff
@@ -453,14 +486,22 @@ <h4 class="modal-title" id="saveModalLabel">Choose a name for your project</h4>
               saveProject(name);
             }
           }
+        }
+        
         },
-      },
```
stevekrouse: It looks like you messed up the formatting of these closing brackets. Please put them back the way it was before.

create.html
```diff
@@ -453,14 +486,22 @@ <h4 class="modal-title" id="saveModalLabel">Choose a name for your project</h4>
...
         error: function(newError) {
           if (newError && newError.lineno) {
             CodeMirror.commands.markGutter(editor, "error", newError.lineno, newError.message)
           }
-        }
+        },
+         selectedTheme: function(newTheme) {
+           var selectedTheme = this.selectedTheme;
+           editor.setOption("theme", this.selectedTheme);
+           if (firebase.auth().onAuthStateChanged) {
```
stevekrouse: Again, this doesn't mean the user is logged in

create.html
```diff
@@ -492,11 +533,11 @@ <h4 class="modal-title" id="saveModalLabel">Choose a name for your project</h4>
     var authChanged = false
     firebase.auth().onAuthStateChanged(function(user) {
       if (!authChanged){
-        authChanged = true
-        if(user){
-          ga('set', 'userId', user.uid); 
-        } 
-      }
+          authChanged = true
```
stevekrouse: Why did you change this code?

create.html
```diff
@@ -511,6 +552,12 @@ <h4 class="modal-title" id="saveModalLabel">Choose a name for your project</h4>
           }
         })
         
+          firebase.database().ref('/user/' + firebase.auth().currentUser.uid + "/theme").once('value').then(function(snapshot) {
```
stevekrouse: Again, 1) this code doesn't only run when the user is logging in and 2) you can combine this firebase request with another firebase request.

create.html
```diff
@@ -511,6 +552,12 @@ <h4 class="modal-title" id="saveModalLabel">Choose a name for your project</h4>
...
+             var selectedTheme = snapshot.val() ;
+             editor.setOption("theme", snapshot.val());
```
stevekrouse: You can simply this line by using the variable you create above

create.html
```diff
@@ -454,13 +485,21 @@ <h4 class="modal-title" id="saveModalLabel">Choose a name for your project</h4>
             }
           }
         },
+        selectTheme: function (){
```
stevekrouse: The name of this method could be more descriptive

create.html
```diff
@@ -454,13 +485,21 @@ <h4 class="modal-title" id="saveModalLabel">Choose a name for your project</h4>
...
+          var yourSelect = document.getElementById( "themeOptions" );
```
stevekrouse: It should be a yellow flag whenever you use `document.getElementById` in your VueJS code. You can get the element here via `event.target`.

You can see how they pass in an `event` argument to a method here as an example: https://vuejs.org/v2/guide/events.html#Methods-in-Inline-Handlers

create.html
```diff
@@ -493,23 +532,25 @@ <h4 class="modal-title" id="saveModalLabel">Choose a name for your project</h4>
...
         authChanged = true
-        if(user){
+          if(user){
           ga('set', 'userId', user.uid); 
         } 
       }
       
       if (user){
         handleLogin(user.displayName, user.uid)
         // if the user is logged in, and we don't have their name/email in our database, save it now
```
stevekrouse: You have modified the code that corresponds to this comment. It used to check if the user exists but now it checks if the user's theme exists in the database.

Thus you have two options:

1) Separate your code entirely from this code so that you don't get its way. The code that was previously here should be the same as what was here. Your new code should be below it.

2) Figure out how to better integrate your new code with the existing code so that it still checks if the user exists (and not just if the user's theme exists) in the database.

create.html
```diff
@@ -326,10 +340,28 @@ <h4 class="modal-title" id="saveModalLabel">Choose a name for your project</h4>
...
       <div v-show="codePane" :style="{display: !codebar ? 'flex' : 'none'}" id="code">
         <div id="CodeNav">
-          <span style="margin-left:8px; margin-top:4px">Code</span>
-          <span style="margin-right:8px">
-            <img @click="codebar = !codebar" style="margin: 5px; width: 16px; height:18px; cursor:pointer" src="./images/minimize.png" title="Minimize">
-          </span>
+           <div>
+            <span style="margin-left:8px; margin-top:4px">Code</span>
+               <select id="themeOptions" style="color:#49A6DF; margin-top: 4px; height: 18px; font-size: 12px;" @change="selectTheme()">
+               <option>monokai</option>
```
stevekrouse: All of these options need to be indented

create.html
```diff
@@ -493,23 +532,25 @@ <h4 class="modal-title" id="saveModalLabel">Choose a name for your project</h4>
...
-          if (!snapshot.val()) {
-            firebase.database().ref("/user/" + user.uid).set({
-              displayName: user.displayName,
-              email: user.email,
-              createdDate: firebase.database.ServerValue.TIMESTAMP
-            })
-          }
-        })
+          firebase.database().ref('/user/' + user.uid + "/theme").once('value').then(function(snapshot) {
+            document.getElementById("themeOptions").value = snapshot.val() ;
```
stevekrouse: It should be a yellow flag whenever you use `document.getElementById` in your VueJS code. The proper way to get the HTML to update in VueJS is to "bind" the HTML to a VueJS data variable. Then, all you have to do in your code is change that variable and the HTML will update automatically. 

However, in this case, VueJS doesn't tell you exactly how to do that with `<select>`. Through trial and error, I figured out that you can do it by adding `:value="selectedTheme"` to your `<select>` element.

All that said, given that VueJS is making it hard to use the `<select>` element without a vue-model, if this proves tricky, you can keep the `document.getElementById` in this case.

create.html
```diff
@@ -501,15 +542,17 @@ <h4 class="modal-title" id="saveModalLabel">Choose a name for your project</h4>
       if (user){
         handleLogin(user.displayName, user.uid)
         // if the user is logged in, and we don't have their name/email in our database, save it now
```
stevekrouse: It looks like you forgot to address this issue:

You have modified the code that corresponds to this comment. It used to check if the user exists but now it checks if the user's theme exists in the database.

Thus you have two options:

Separate your code entirely from this code so that you don't get its way. The code that was previously here should be the same as what was here. Your new code should be below it.

Figure out how to better integrate your new code with the existing code so that it still checks if the user exists (and not just if the user's theme exists) in the database.

Issue #455, opened 2017-07-25
Migrate/update old users


jwass91:
With the new sign up that requires a unique username and the option to receive the newsletter, all old users need to be updated. 

- Whenever a user logs in we will check if they are "new" or "old"
- To check if a user is "old" we just need to see if the newsletter field exists for them
- Then if they are an "old" user we will direct them to some sort of forms that requires them to choose a new username (or possibly keep their old one) and check off if they want to receive the newsletter
-  Then we will update the users info and add their new username to the unique username list

stevekrouse:
Sounds like a great plan!

Issue #451, opened 2017-07-25
ensure usernames uniqueness is consistent in the database


stevekrouse:
There's a small but non-zero chance that between when you check if the unique username exists in the database and when you store the unique username in the database, another user is creating that exact same username and beats you to it.

The solution here is that instead of querying the uniqueusername field, we try to update it first. If it succeeds, we're good and if it fails, we know that someone already has that username. We'll need to make sure to set the rules in firebase to not allow you to set a username field that already exists.

PR #458, opened 2017-07-27
Migrate Old Users

jwass91:
addresses #455

### DIFF
--- a/create.html
+++ b/create.html
@@ -863,8 +863,6 @@
         .then(function() {
           $('#loginModal').modal('hide');
           clearLoginSignupModal();
-
-          user = firebase.auth().currentUser
           app.projects = []
 
         })
@@ -1003,6 +1001,7 @@
 
                 
                 app.userNameForMenu = app.signUpUsername
+                app.projects = []
 
                 firebase.database().ref("uniqueUsername/" + app.signUpUsername).set({
                   username: app.signUpUsername,
--- a/create.html
+++ b/create.html
@@ -305,6 +305,37 @@
       </div>
     </div>
 
+    <div class="modal fade" id="migrateModal" tabindex="-1" role="dialog" aria-labelledby="migrateModal">
+      <div class="modal-dialog" role="document">
+        <div class="modal-content">
+          <div class="modal-header">
+            <h4 class="modal-title" id="migrateModalLabel">We are updating our systems and need you to choose a new username. Don't use your real name!</h4>
+          </div>
+          <div class="modal-body">
+            <div class="form-group" :class="{'has-success': migrateUsernameValid == 'VALID', 'has-error': migrateUsernameValid == 'TAKEN', animated: migrateUsernameValid == 'TAKEN' || migrateUsernameKeyIllegal,  shake: migrateUsernameValid == 'TAKEN' || migrateUsernameKeyIllegal}">
+              <label for='username'>Username</label>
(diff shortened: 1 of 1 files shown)


stevekrouse:

https://cdn.rawgit.com/stevekrouse/WoofJS/41d7a2786aeb8679d127f14d5c051957c16b8e3b/index.html

create.html
```diff
@@ -305,6 +305,37 @@ <h4 class="modal-title" id="saveModalLabel">Choose a name for your project</h4>
...
+              <label for='username'>Username</label>
+              <input v-model='migrateUsername' onfocus="$('#usernameMigrate').tooltip('toggle')" data-placement="top" title="Don't use your real name" type="text" class="form-control" id="usernameMigrate" placeholder="Username" aria-describedby="usernameMigrateHelp" @input="migrateUsernameVaidate($event)" onkeydown="if (event.keyCode == 13){migrate()}" >
+              <span id="usernameMigrateHelp" class="help-block" :style="{display: migrateUsernameValid == 'TAKEN' ? 'block' : 'none'}">This username is already taken. Please try another username.</span>
+              <span id="usernaemeMigrateHelp" class="help-block" style='color:#BD362F;' :style="{display: migrateUsernameKeyIllegal ? 'block' : 'none'}">You cannot include {{ migrateUsernameKeyIllegal }} in usernames.</span>
+            </div>
+            <div class="checkbox">
+              <label>
+                <input type="checkbox" v-model='migrateChecked'> I want to recieve updates about WoofJS
```
stevekrouse: Where's the closing tag on this `<input>`?

create.html
```diff
@@ -806,6 +863,10 @@ <h4 class="modal-body text-danger" :style="{display: !getID() ? 'block' : 'none'
         .then(function() {
           $('#loginModal').modal('hide');
           clearLoginSignupModal();
+
+          user = firebase.auth().currentUser
```
stevekrouse: What does setting user here accomplish?

create.html
```diff
@@ -806,6 +863,10 @@ <h4 class="modal-body text-danger" :style="{display: !getID() ? 'block' : 'none'
...
+          app.projects = []
```
stevekrouse: Good call!

create.html
```diff
@@ -821,6 +882,64 @@ <h4 class="modal-body text-danger" :style="{display: !getID() ? 'block' : 'none'
...
     }
 
+    function migrate(){
+      if (app.migrateUsernameValid == 'VALID' && app.migrateButtonState != 'LOADING') {
+        app.migrateButtonState = "LOADING"
+
+        var exists
+        var ref = firebase.database().ref("uniqueUsername");
+
+        ref.child(app.migrateUsername).once('value', function(snapshot) {
```
stevekrouse: We have a potential consistency issue here like before... https://github.com/stevekrouse/WoofJS/issues/451

Again, no need to solve it now, but just a heads up that it will need to be solved eventually.

create.html
jwass91: I thought you don't need one. The "I want..." message is not a part of the input so doesn't need to be enclosed by an input tag.

create.html
jwass91: I had it there to help reset the user menu. Before there were some bugs with the name not changing and the old user's project remaining there.

stevekrouse: Totally, I'm with you. I just don't see which global variable you're affecting here. It looks to me like you're setting a local variable which woudn't have any effect anywhere else in the code. For contrast, it's very clear to me how setting `app.projects` here will change things. Can you point me to where this `user` variable is defined so it's clear to me what you're modifying?

stevekrouse:
Feel free to merge when you're ready. My comment on setting the `user` variable is just a curiosity. Won't affect things really, I'm just confused.

create.html
jwass91: I think I can probably just get rid of it and nothing will change. Not sure exactly what I was doing when I put it there.

stevekrouse:
Looks good to me! Feel free to merge if you're ready :)

PR #459, opened 2017-07-28
Login with username

jwass91:
@stevekrouse 
There is an issue here with the error messages. 

It is hard to determine what the error messages should be so that they are helpful while not compromising security. Due to the fact that multiple users could have the same username, we need some sort of error message to indicate that then you have to sign in with your email address instead. 

Right now the errors are:
- "No account with this username or email exists. Please try to sign up for an account."
- "Invalid email address. Please try again."
- "Multiple users share this username, please login with your email."


stevekrouse:

https://cdn.rawgit.com/stevekrouse/WoofJS/67ef155c01ab7f976be53c8873db4f839d5a9320/index.html

stevekrouse:
This looks great!

My thought was that we should only allow people who have unique usernames (which we have validated because they are in the `uniqueUsername` database) to sign in with their username. In fact, I was thinking that you could look up these users via the `uniqueUsername` database (it's a simple get call, no sorting necessary). 

For the users without a unique username, they have to login via email. (There won't be many of these as we are migrating users via your other pull request)

stevekrouse:
Looks great! Feel free to merge when you're ready.

As mentioned above, my preference would be to do the lookup in the `uniqueusername` database instead but this seems to work great, so whatever you think is best.

jwass91:
Merged into migrateUsers. Closed in favor of #458.

Issue #328, opened 2017-03-08
Add tests for woofjs core library


stevekrouse:
There are a lot of quick sanity checks that we can add to woof.js to ensure that we didn't accidentally break something. For example, we can test that when you move a sprite, that it's x and y position change accordingly. 

One interesting test we can run is to make sure that `babel(src/woof.es6.js) === dist/woof.js` because this is a common bug that I make all the time. We can then instruct developers to add a pre-commit hook that checks this. 

Open questions

1. What JS testing framework should we use?

2. Would it make things easier to run the tests without a DOM? We can always mock out a DOM or do something clever so that rendering doesn't happen when we're running the tests.

3. How can we get the tests to run on each push to master? Can we get the tests to run on each push to a branch / pull request? Which testing service will we use?

stevekrouse:
The tests should also make sure there aren't any postimage or imgur links in the codebase

Issue #326, opened 2017-03-07
add package.json


stevekrouse:
We currently don't have a `package.json` file in woof. Because of this, we actually end up asking developers to create one in the readme with `npm init --yes` and then install babel manually.

Instead of this, we should have a `package.json` file in woof so they don't have to init npm. Then can just say `npm install` or whatever the command is and it will automatically go grab babel.

stevekrouse:
Hopefully this would also include a command to run the tests. 

And a command to compile all the files. In theory this would allow us to slowly add a more robust build system to the project although I do like how simple things are now.

stevekrouse:
We could also add a `.gitignore`

joebeachjoebeach:
I came to post this idea as a new issue myself, and lo and behold, it's been here all along.

It would really streamline the development process, and allow new contributors to get up and running more quickly. I'm happy to take it on.

In terms of adding to the build system, we could use <a href="https://www.npmjs.com/package/prettier" target="_blank">prettier</a> to standardize code formatting, which would eliminate the need for code revisions/edits just for formatting.

stevekrouse:
Agreed! This is also somewhat related to #328. 

I'd love the idea of having prettier standardize code formatting. Give it a shot and see what happens. I imagine this would be a pretty big pull request because you'd be reformatting every line of every file.

joebeachjoebeach:
Cool! I'm excited about this.

Do you have any adjustments you'd like to make to the [default prettier rules](https://github.com/prettier/prettier#options)?

stevekrouse:
I wonder if we can somehow integrate it into c9...

I think you'll have to play with the print width.

Semicolons should be set to false.

Trailing commas "es5"

Bracket spacing false

joebeachjoebeach:
I'll look into c9 integration. I personally have decided to stop using c9 for Woof development because it's been unreliable for me for whatever reason, so I'd like to have an environment-agnostic solution.

Another suggestion for further down the road: what would you think of breaking the codebase into modules and compiling with Webpack? We could break things up into something like `main/`, `sprites/`, `control/`, `math/`, and `helpers/`. The Webpack compiling could be bundled in with whatever npm script we use for Babel compiling. Would this be worth the effort or is it needlessly complicating things?

stevekrouse:
In my experience webpack and other build tools are a pain in the ass. I really like how dead simple the WoofJS build experience is: basically none. 

[...]

PR #466, opened 2017-08-14
start adding npm packages and scripts

joebeachjoebeach:
Addresses #326 

This is just a first step, but I wanted to check some things first.

I added Babel and Prettier, and I added the following scripts for them:
- `npm run compile` : compile src/woof.es6.js to dist/woof.js
- `npm run watch` : same as above but with the `--watch` flag to re-compile on every change
- `npm run format` : format src/woof.es6.js with prettier

I only added src/woof.es6.js to prettier for now just to test it out. I'm guessing we'd want it to format all *.js files if we decide to go forward with it, right?

Some issues with Prettier: 
- it formats template literal strings a little oddly. You can see this evidenced on line 625 of src/woof.es6.js (they have an open issue about this [here](https://github.com/prettier/prettier/issues/1893))
- turning off semicolon insertion still results in some semicolons being inserted at the beginning of lines, which is heavily evident in the SAT.js code at the top of src/woof.es6.js (this is mentioned in the [docs](https://github.com/prettier/prettier#semicolons))

That being said, some questions for @stevekrouse:

1. Any comments on those npm scripts?
2. Should we go ahead with prettier?
3. Do we want to format all *.js files (excluding node_modules/ of course, and probably excluding everything in vendor/codemirror/)?
4. What do you think of adding the prettier script as a pre-commit hook as referenced in the [prettier docs](https://github.com/prettier/prettier#pre-commit-hook)? It'd be nice if we could do the same with the babel script, but I'm not positive we can. (that would sort-of address one of the aspects of #328)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,9 @@
 # Ignore Mac DS_Store files
 .DS_Store
+
+# Ignore any sublime project files
+*.sublime-project
+*.sublime-workspace
+
+# Ignore node_modules folder
+node_modules
--- a/.gitignore
+++ b/.gitignore
@@ -5,5 +5,8 @@
 *.sublime-project
 *.sublime-workspace
 
+# Ignore any eslint config files
+.eslintrc.json
+
 # Ignore node_modules folder
 node_modules
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,6 @@
 
 # Ignore node_modules folder
 node_modules
+
+# Ignore any eslint config files
+.eslintrc.json
--- a/.gitignore
+++ b/.gitignore
@@ -5,8 +5,5 @@
 *.sublime-project
 *.sublime-workspace
 
-# Ignore any eslint config files
-.eslintrc.json
-
 # Ignore node_modules folder
 node_modules
--- a/README.md
+++ b/README.md
@@ -112,20 +112,13 @@ The documentation that you find on [woofjs.com/create.html](http://woofjs.com/cr
 
 #### woof.js core library
 
-If you'd like to contribute to the core WoofJS library, you will edit the code in src/woofjs.es6.js , but these edits need to be compiled to dist/woof.js using babel
+If you'd like to contribute to the core WoofJS library, you will edit the code in src/woof.es6.js , but these edits need to be compiled to dist/woof.js using babel
 
-There are two options to do this:
-
-1. Every time you make changes and want to implement them, enter this command to compile:
-
-    ```
-    ./node_modules/.bin/babel WoofJS/src/woof.es6.js --out-file WoofJS/dist/woof.js
-    ```
-2. Or -- tell babel to always watch for changes and auto-compile (less work for you!). Open a new terminal and run this:
-
-    ```
-    ./node_modules/.bin/babel WoofJS/src/woof.es6.js --watch --out-file WoofJS/dist/woof.js
-    ```
+To do that, make sure you've installed the npm dependencies with `npm install`, then, before you start editing files, run the following command to tell babel to watch for changes and automatically compile:
+   ```
+   npm run compile
+   ```
+You can abort this process at any time by pressing **CTRL+C**
 
 #### WoofJS.com's Tutorials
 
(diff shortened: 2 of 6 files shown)


stevekrouse:

https://cdn.rawgit.com/joebeachjoebeach/WoofJS/43fc6581afd1a69796f8020b7d970911a56e2764/index.html

stevekrouse:
Thanks for taking the lead here @joebeachjoebeach and thinking through these issues!

While I do love the idea of a formatter, it seems like a big project to move all our code to that style. Maybe we can make that a separate project and leave this one to just to compile and watch?

Another question for you -- how did you produce the package-lock file? That seems really long and complex, especially considering that our use-case is so simple.

As far as the API to run these commands, it seems to me like there's really just one command `babel run compile` and you can optionally give it the watch flag `babel run compile --watch` if you want. I think if we're doing prettier than it should automatically prettify the code in the compile command as well (but we can leave that for a seperate issue/pull request). I think that might make more sense than a pre-commit hook

joebeachjoebeach:
Roger that re: prettier. I'll take that out and leave this to just npm + babel.

package-lock.json is generated by npm; it's a relatively new thing. [This](https://stackoverflow.com/questions/44297803/package-lock-json-role) can provide more info than I'm capable of giving. It's intended to be committed, but let me know if you'd rather I add it to .gitignore.

I'm in favor of just having the one command to watch the file, aborting the operation with ctl-c at will. Did you mean `npm run compile` though? I think we have to make it an npm script unless I'm missing something.

And I'll definitely thoroughly update the docs once everything's final.

stevekrouse:
Good to know about `package-lock.json`, sounds great to add it like you did in this pull.

Yep, `npm run compile` is what I'm thinking.

joebeachjoebeach:
OK, I update the script to be just `npm run compile` and added the necessary instructions to the readme.

Did the readme once have instructions to fork the repo then clone, or is that a dream that I made up? Do we want to put those instructions in there to promote best practices and avoid someone accidentally committing the `master` branch?

joebeachjoebeach:
Additionally, when this is merged, will it potentially interfere with some contributors' development workflow? They may have to re-install the npm dependencies, or their existing node_modules folder may be in a parent directory. Should we send out a message to all the contributors to alert them to this change?

.gitignore
```diff
@@ -1,2 +1,12 @@
...
+
+# Ignore any sublime project files
+*.sublime-project
+*.sublime-workspace
+
+# Ignore node_modules folder
+node_modules
+
+# Ignore any eslint config files
+.eslintrc.json
```
stevekrouse: Seems like this would be a great thing to commit, no?

README.md
```diff
@@ -142,34 +135,29 @@ You can find them there in the `woof/` subdirectory. If you'd like to cotribute
...
 
     ```
-    npm install babel-cli babel-preset-es2015 --save-dev
+    npm install
     ```
-5. Take a look at our issues and see if there are any issues that speak to you as good places to tackle first. Pay particular attention to issues labeled "good student projects."
+4. Take a look at our issues and see if there are any issues that speak to you as good places to tackle first. Pay particular attention to issues labeled "good student projects."
 
-6. Create a new local branch to work on your issue -- use a descriptive branch name like in this example:
+5. Create a new local branch to work on your issue -- use a descriptive branch name like in this example:
```
stevekrouse: This makes me realize that we should probably work on this (in a separate pull req): https://github.com/stevekrouse/WoofJS/issues/486

stevekrouse:
When I pulled this branch, `npm run compile` worked great for me without running `npm install` which makes sense because it's the same command and I had that stuff installed already, so I imagine that'd be true for other developers.

I spun up a new c9 instance and pulled your branch clean and ran the steps and then went perfectly. Great work!

.gitignore
joebeachjoebeach: I added it so that it would ignore my personal .eslintrc file for now. Since c9.io uses eslint by default, I didn't want my preferred linting rules interfering with others' preferences without warning/discussion. I do think it would be a good idea to come up with a standard set of linting rules, though. I'll create a separate issue for this.

README.md
joebeachjoebeach: Oh, interesting. I wasn't aware of the distinction, but it makes sense, I suppose.

Issue #396, opened 2017-06-09
Variable misspelled. Do you mean...?


stevekrouse:
I wonder if there's a library we can use that will help us provide suggestions for students when they get an undefined variable error when it's a simple misspelling.

stevekrouse:
This or a similar library might be helpful: https://github.com/hiddentao/fast-levenshtein

You could get the text of the code, split on space " ", and then compute the edit distance of the mispelled variable to each of those terms. If it's less than 2 or 3, suggest it. 

Additionally, you could throw a list of WoofJS methods and terms (penDown, Image, etc) into the list that you compute the edit distance from so you can offer those as suggestions for misspelled words. Because although this issue says "variable misspelled" it technically be for an identifier misspelled, including functions.

Issue #249, opened 2016-11-11
autocomplete on variable names


stevekrouse:
Using something like this potentially: https://codemirror.net/demo/anywordhint.html

But hopefully in a way that doesn't double-add forever() and all other non-variable words to the list

PR #468, opened 2017-08-15
add suggestions for misspelled variables and methods

madduccino:
Addresses #396 - Offers suggestions for misspelled variables and Woof methods using Levenshtein library if distance between words is less than 3.

### DIFF
--- a/create.html
+++ b/create.html
@@ -650,7 +650,6 @@
             this.referenceError = message.split(" ")[0].split(".").pop();
             // Get alphanumeric user code
             var userCode = editor.getValue().replace(/(?:https?|ftp):\/\/[\n\S]+/g,'').match(/[a-zA-Z0-9]+/g);
-            console.log(userCode)
             // Add Woof methods to the user code 
             userCode.push("onKeyDown", "onMouseDown", "onMouseUp", "onMouseMove", "penDown", "Image", "Rectangle", "Polygon", "Circle", "Line", "Oval", "move", "turnRight", "turnLeft", "angle", "pointTowards", "setRotationStyle", "show", "hide", "sendToFront", "sendToBack", "play", "Audio", "clearPen", "penUp", "penDown", "Text", "remove", "includes", "forEach", "find", "setBackdropURL", "setBackdropStyle", "after", "every", "repeat", "forever", "if", "else", "mouseX", "mouseY", "repeatUntil", "when", "freeze", "defrost", "delete", "remove", "push", "throttle", "mouseOver", "randomX", "randomY", "touching", "distanceTo", "prompt", "alert", "keysDown", "mouseDown", "pMouseX", "pMouseY", "mouseXSpeed", "mouseYSpeed", "hour", "hourMilitary", "minute", "second", "dayOfMonth", "dayOfWeek", "month", "year", "mobile", "random", "randomColor", "between", "substring", "round", "sqrt", "abs", "floor", "ceiling", "importCodeURL")
             // Search user code for suggestions 
--- a/create.html
+++ b/create.html
@@ -649,13 +649,14 @@
             // Isolate the text causing the error
             this.referenceError = message.split(" ")[0].split(".").pop();
             // Get alphanumeric user code
-            var userCode = editor.getValue().match(/[a-zA-Z]+/g);
+            var userCode = editor.getValue().replace(/(?:https?|ftp):\/\/[\n\S]+/g,'').match(/[a-zA-Z0-9]+/g);
+            console.log(userCode)
             // Add Woof methods to the user code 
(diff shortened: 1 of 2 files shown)


stevekrouse:

https://cdn.rawgit.com/stevekrouse/WoofJS/80807102cea0ad444cb69645453a791e2612cc31/index.html

stevekrouse:
Really amazing work @madduccino!

I have a few ideas here:

1) I really love the error div at the top of the code pane! I think we can remove the error button entirely and just have all error messages display there when they happen.  What do you think of that?

2) Your code only runs when the `showError` button in clicked. However, you probably want this code to run whenever the `app.error` value changes. You can accomplish this with the vue watch functionality: https://vuejs.org/v2/guide/computed.html#Watchers

create.html
```diff
@@ -601,20 +634,46 @@ <h4 class="modal-body text-danger" :style="{display: !getID() ? 'block' : 'none'
...
+            // Search user code for suggestions 
+            for (var i = 0; i < userCode.length; i++) {
+              var distance = window.Levenshtein.get(this.referenceError, userCode[i]);
+              if (distance < 3 && this.suggestions.indexOf(userCode[i]) == -1 && userCode[i].length > 1) {
```
stevekrouse: You could add a comment here like you did above

create.html
```diff
@@ -601,20 +634,46 @@ <h4 class="modal-body text-danger" :style="{display: !getID() ? 'block' : 'none'
...
+                this.suggestions.push(userCode[i]);
+                if (distance < 1) {
+                  this.suggestions.splice(this.suggestions.indexOf(userCode[i], 1));
```
stevekrouse: comment here could be nice as well

create.html
```diff
@@ -601,20 +634,46 @@ <h4 class="modal-body text-danger" :style="{display: !getID() ? 'block' : 'none'
...
           }
           else {
-             alert("Error on line " + this.error.lineno  + "\n\n" + this.error.message)
+            alert("Error on line " + this.error.lineno  + "\n\n" + this.error.message)
           }
         },
+        // Change the user code if user clicks suggestion
+        takeSuggestion: function(x) {
+          editor.setValue(editor.getValue().replace(new RegExp("\\b" + this.referenceError + "\\b", 'g'), x));
+          document.getElementById("suggestVar").style.display = "none";
```
stevekrouse: Usually, using `document.getElementById` is a sign that you're not leveraging VueJS enough. Why don't you just set `this.suggestVar = false`?

create.html
```diff
@@ -748,7 +807,22 @@ <h4 class="modal-body text-danger" :style="{display: !getID() ? 'block' : 'none'
         }
       }
     })
-  
+
+    // Toggle misspelled variable suggestion
+    $(document).ready(function(){
+      $("#error").click(function(){
```
stevekrouse: Why are you hiding and showing things manually here? Again, VueJS should take care of this for you, especially because you already created a v-show for this.

create.html
```diff
@@ -601,20 +634,46 @@ <h4 class="modal-body text-danger" :style="{display: !getID() ? 'block' : 'none'
...
           if (this.error.message.includes('Uncaught ReferenceError: sprite1 is not defined')) {
             var spriteName = prompt("Error on line " + this.error.lineno  + "\n\n" + "You need to rename 'sprite1'. What would you like to name it?", "unicornSprite");
             if (spriteName !== null) {
                 editor.setValue(editor.getValue().replace(/sprite1/g, spriteName));
             }
           }
-          else if (this.error.message.includes("Uncaught ReferenceError")) {
-            var message = this.error.message.replace("Uncaught ReferenceError: ", "")
-            alert("Error on line " + this.error.lineno  + "\n\n" + message + ". Check the spelling of your variable definitions.")
+          else if (this.error.message.includes("Uncaught ReferenceError") ||
```
stevekrouse: This code probably should run whenever there's an error, not just when the error button is clicked. However, given that we're getting rid of the button (probably), that shouldn't be as much of a problem :)

madduccino:
Removed the error button plus other changes

madduccino:
Also, I've noticed that line numbers for errors defined in the WoofJS library are wonky.

screenshot 2017-08-27 at 7 05 33 pm

stevekrouse:
Agreed @madduccino! https://github.com/stevekrouse/WoofJS/issues/483

stevekrouse:
Now that our error messages are getting really really good, I had another idea: https://github.com/stevekrouse/WoofJS/issues/484

stevekrouse:
It also makes me think that this issue should be top-priority: https://github.com/stevekrouse/WoofJS/issues/249

stevekrouse:
Another idea! https://github.com/stevekrouse/WoofJS/issues/485

stevekrouse:
This is amazing work, @madduccino! So pumped to get this into production! This work is so good that you've inspired me to create a whole bunch of new issues (see above) in this vein.

PR #472, opened 2017-08-19
fixed spaces in project names bug

jwass91:
I noticed that the project naming had a bug that allowed for spaces in project names. If you copied and pasted a string with several spaces it would only replace the first space. This fixes that issue by using a method that replaces all spaces in the string.

### DIFF
--- a/create.html
+++ b/create.html
@@ -699,7 +699,7 @@
           }.bind(this), 10)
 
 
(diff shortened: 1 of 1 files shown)


stevekrouse:

https://cdn.rawgit.com/stevekrouse/WoofJS/7bf4b10ce8a3024474a9142cb7868c1a22b5b71d/index.html

PR #481, opened 2017-08-31
Project meta data

stevekrouse:
Before we merge this in, it probably makes sense to delete the `code-meta-data` table and rerun the migration in the code below.

### DIFF
--- a/analytics.html
+++ b/analytics.html
@@ -38,8 +38,6 @@
 </style>
 <div id="page" style="display: none" :style="{display: authenticated ? 'block' : 'none'}">
     <h1>WoofJS.com Analytics</h1>
-    <div style="height: 350px; width: 100%" id="chart"></div>
-    <div style="width: 100%; text-align: center; margin-bottom: 20px" id="legend"></div>
     <div id="totals" class="tablewrap">
         <h2>Totals</h2>
         <analytics-table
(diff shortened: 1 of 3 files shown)


stevekrouse:
Hey, @jwass91 can you please review this, giving me feedback where you have it and asking questions where confused? Double check for me that this schema will work for your social network use-cases

stevekrouse:

https://cdn.rawgit.com/stevekrouse/WoofJS/3583be06df896ce87cddc504feb190ec520ee0bb/index.html

stevekrouse:
One thing I forgot from this pull request that we'll probably want: `projectCreatedTime` at the top level of a `code-meta-data` project so 1) we don't have to calculate it by looking at the original revision and 2) so we can sort by it  in firebase.

Additionally, we'll probably want to verify the migration worked well with a test that makes sure the schema is good. This can either be solved by pulling the data after the migration and running tests on it or by adding schema tests on the data on the firebase side. The second way is probably the way to go. We can probably also do this with the other data tables. For example,I lost a lot of time yesterday debugging an issue where one of the entries in the `user` firebase table was `displayName: "jwass91"` which was annoying and a better schema rule would've prevented that in the first place.

analytics.html
```diff
@@ -136,31 +134,35 @@
...
   
   var utils = {}
-  utils.projectFirstRevision = function(project){
-      return project.revisions[0]
+  utils.projectFirstRevisionTime = function(project){
+    if (project.revisions.length > 0) {
+      return project.revisions[0].time
+    } else {
+      return project.currentVersionTime
+    }
```
jwass91: We should add projectFirstRevisionTime as first level meta data so this will not be necessary.

analytics.html
```diff
@@ -136,31 +134,35 @@
...
   }
-  utils.projectLastRevision = function(project){
-      return project.revisions[project.revisions.length - 1]
+  utils.projectLastRevisionTime = function(project){
+    return project.currentVersionTime
   }
```
jwass91: Why do you need this as a util function? Couldn't you just use project.currentVersionTime.

create.html
```diff
@@ -1211,37 +1204,42 @@ <h4 class="modal-body text-danger" :style="{display: !getID() ? 'block' : 'none'
...
+      updates["/code-meta-data/"+ name + "/version/" + revisionId] = firebase.database.ServerValue.TIMESTAMP
+      
+      firebase.database().ref().update(updates).then(function(){
+        window.location.hash = name
+        myCode = firebase.auth().currentUser
+        makeClean()
+      }).catch(function(error){
+        alert("Unable to save project: " + error)
+      });
+    }
```
jwass91: We should add the projectFirstRevisionTime at the top level of the meta data.

firebase-migrations.js
```diff
@@ -0,0 +1,63 @@
...
+        if (revision.time) {
+          updates["/code-meta-data/"+ project.name + "/version/" + revision.id] = revision.time
+        }
+      })
+    }
+  })
+  
+  firebase.database().ref().update(updates)
+})
+
```
jwass91: Looks good. We probably will want to run this again right after merging to migrate the projects created between the first run and the merge.

jwass91:
Can we also add displayName at the top level of the meta data so we don't need extra queries to go from uid to displayName. For anonymous projects, we can just not have that field.

Issue #485, opened 2017-09-01
Error line number should be a link


stevekrouse:
When you click on it, it should scroll to that line and maybe also highlight that line

Issue #483, opened 2017-09-01
Better line numbers on WoofJS method input errors


stevekrouse:
![image](https://user-images.githubusercontent.com/2288939/29973373-56dd6ce4-8efd-11e7-8e4a-b81c8be4b3ba.png)

hallpell:
These types of errors no longer include a line number (there isn't any indication where the error comes from, which isn't great, but is at least less confusing).

PR #492, opened 2017-09-16
add error line links

madduccino:
Addresses issue #485 by doing the following:

- Adds link to error line in editor
- Highlights line when user clicks link
- Extracts line error number from error message when line number is undefined
- Hides error number when it is outside the range of the editor (until resolved in issue #483)


stevekrouse:

https://cdn.rawgit.com/stevekrouse/WoofJS/492f97aee6ee2b12bfa01111b239e2e48e8ec8b5/index.html

create.html
```diff
@@ -654,6 +656,9 @@ <h4 class="modal-body text-danger" :style="{display: !getID() ? 'block' : 'none'
         suggestions: [],
         suggestVar: false,
         errorMessage: '',
+        errorLineInfo: '',
+        currErrorLineNo: '',
+        knowErrorLineNo: false,
```
stevekrouse: It seems like these three variables could be collapsed into one: `errorLineNumber` (I like this better than `currErrorLineNo` but I don't really care that much if you prefer what you have). It is initially set to `undefined` (instead of the empty string which makes less sense because this variable is never a string, but always either a number or undefined) which would be the equivalent of `knowErrorLineNo == false` (and whenever it is set to a number, that implies `knowErrorLineNo == true`).

`errorLineInfo`'s text can be built within the HTML on line 613. No need to do it on line 752. The way I think about it is that your core data code is constructed in JS, and the visual layer is in HTML as a function of that core data.

create.html
```diff
@@ -729,19 +731,28 @@ <h4 class="modal-body text-danger" :style="{display: !getID() ? 'block' : 'none'
...
+            // Find error line number if it is undefined
+            if (this.error.lineno == undefined) {
+              var x = errorDesc[0].slice(errorDesc[0].indexOf(":") + 2);
+              this.error.lineno = x.slice(x.indexOf("(") + 1, x.indexOf(":"));
+            }
+          }
+          // Show error line number if it is defined and line number is accurate
+          if (this.error.lineno != undefined && this.error.lineno <= editor.lineCount()) {
+            this.knowErrorLineNo = true
+            this.errorLineInfo = "Error on Line " + this.error.lineno + ":";
```
stevekrouse: I have an idea about how to improve this text, which I explain in a comment above should be in the HTML (except maybe a function for the random emoji). Here's what I got when I monkeyed around with it:

screenshot 2017-09-19 at 2 15 01 pm

Among other things, I like begin the error message with a sad emoji -- kinda fun, no? I thought it'd be fun if it picked randomly from one of the following options: 😒 😞 😔 😟 😕 🙁 ☹️ 😣 😖 😫 😩. I'm sure you could find other good options too if you looked :)

create.html
```diff
@@ -605,15 +610,12 @@ <h4 class="modal-body text-danger" :style="{display: !getID() ? 'block' : 'none'
...
+        <div id="error-block" :style="{display: seenError ? 'block' : 'none'}" class="alert alert-danger fade in"><a :style="{display: knowErrorLineNo ? 'inline' : 'none'}" @click="goToErrorLine">{{errorLineInfo}}</a>{{errorMessage}}
           <div :style="{display: renameSprite ? 'block' : 'none'}">
             <p>You need to rename sprite1. What would you like to name it?</p> 
             <input @keyup.enter="getSpriteName" class="form-control" type="text" v-model="spriteName" placeholder="Enter sprite name"><button class="btn btn-danger" @click="getSpriteName">Enter</button>
           </div>
-          <ul :style="{display: suggestions.length > 0 ? 'block' : 'none'}">
-            <li>Did you mean: <a v-for="suggestion in suggestions" @click="takeSuggestion(suggestion)">{{suggestion}}</a>
-            </li>
-          </ul>
+          <div :style="{display: suggestions.length > 0 ? 'block' : 'none'}">Did you mean: <a v-for="suggestion in suggestions" @click="takeSuggestion(suggestion)">{{suggestion}}</a></div>
```
stevekrouse: I think it'd be neat to wrap each suggestion in `<code></code>` tags.

create.html
```diff
@@ -605,15 +610,12 @@ <h4 class="modal-body text-danger" :style="{display: !getID() ? 'block' : 'none'
             </span>
           </div>
         </div>
-        <div id="error-block" :style="{display: seenError ? 'block' : 'none'}" class="alert alert-danger fade in">{{errorMessage}}
+        <div id="error-block" :style="{display: seenError ? 'block' : 'none'}" class="alert alert-danger fade in"><a :style="{display: knowErrorLineNo ? 'inline' : 'none'}" @click="goToErrorLine">{{errorLineInfo}}</a>{{errorMessage}}
```
stevekrouse: I think it'd be neat to wrap the error message in `<code></code>` tags. I have other ideas for how this could look in a comment below.

create.html
stevekrouse: I also don't love that "Did you mean" is a question without a question mark, but I understand that there's no real place for the question mark that makes sense. Maybe we can replace this with: "You may be looking for:"

create.html
```diff
@@ -269,21 +268,27 @@
...
-    #error-block li a {
-      padding-right: 5px; 
-      cursor: pointer;
+    #error-block div a {
+      padding: 0 7px 0 3px;
     }
     
+    .highlight {
+      color: #f2dede;
+      background-color: #a94442;
```
stevekrouse: Maybe make the color of the line highlight in the text editor lighter. I thought it looked neat when I made it the same color as the error box red `#f2dede` but that's up to you. At the very least a little lighter would be good.

stevekrouse: I also wonder if we should always highlight the error line when we have it automatically. To be clear, I don't mean scroll it into view -- that we can wait on their click. I just mean highlighting it visually. Or do you prefer waiting on users to click on it?

madduccino:
I'm ready for review! (Sorry about all the commits.) I ran into some issues on this:

1. It seemed impossible to find a highlight color that was consistent with the overall error style that also made the cursor visible in every editor theme. My solution here is to add a cursorActivity function that checks whether the active cursor is on the error line. HOWEVER, I have since realized that this makes the randomEmoji function fire every time you move in and out of the error line. Not sure the best way to move forward.

2. With the lighter highlight color, most themes' selected text became no longer visible. So, I incorporated Codemirror's mark-text.js addon which allowed me to change the style of selected text. I probably could have done this on the individual theme style sheets, but I wanted this to be scalable.

I also removed the word "unkown:" should it appear in the error message.

madduccino:
randomEmoji also is triggered each time I enter a character in the rename sprite field ☹️

madduccino:
I solved the randomEmoji problem by making it a variable inside the showError function as opposed to a Vue method.

create.html
```diff
@@ -654,6 +678,10 @@ <h4 class="modal-body text-danger" :style="{display: !getID() ? 'block' : 'none'
         suggestions: [],
         suggestVar: false,
         errorMessage: '',
+        changeCursor: false,
```
stevekrouse: I'd probably rename this variable to `cursorOnErrorLine`

create.html
```diff
@@ -1410,7 +1457,28 @@ <h4 class="modal-body text-danger" :style="{display: !getID() ? 'block' : 'none'
             "asi": true
           }
         },
+        styleSelectedText: true,
```
stevekrouse: It's unclear to me where in the code you "select" the text on the line where the error is occurring...? Could you point that out? It might make sense to add a comment above this line to make its purpose clear

stevekrouse:
Really fun! Love it. Just have two little stylistic comments on your code :)

nicolekelner:
http://www.makeuseof.com/tag/emoji-english-dictionary-emoji-faces-meaning-explained/ Here are apple emoji's that I think would be even cuter! They're all pngs

madduccino:
-Remove seenError
-Refactor watcher into update and show error method

create.html
madduccino: I added this so that if users select the error line (to delete it, comment it, etc.), they can see that it is selected.

This text is selected, but you can't see it:
screenshot 2017-10-01 at 8 29 15 am

The addon makes it possible to style the selected text so you can now see that the text is selected:
screenshot 2017-10-01 at 8 30 21 am

madduccino:
- Swapped emojis
- Refactored error watcher into `showError()` and `tryRunningCode()` functions
- Removed `seenError`
- Replaced `referenceError` with a `findMisspelledWord` method

create.html
```diff
@@ -700,60 +725,73 @@ <h4 class="modal-body text-danger" :style="{display: !getID() ? 'block' : 'none'
...
+          return typo.split(" ")[0].split(".").pop();
+        },
+        // Change user code if user clicks suggestion
         takeSuggestion: function(x) {
-          editor.setValue(editor.getValue().replace(new RegExp("\\b" + this.referenceError + "\\b", 'g'), x));
-          this.suggestVar = false;
+          editor.setValue(editor.getValue().replace(new RegExp("\\b" + this.findMisspelledWord(this.errorMessage) + "\\b", 'g'), x));
         },
         // Rename sprite1 with user input
         getSpriteName: function() {
```
stevekrouse: The name of this function doesn't seem quite right

create.html
```diff
@@ -1474,55 +1520,50 @@ <h4 class="modal-body text-danger" :style="{display: !getID() ? 'block' : 'none'
...
-              app.mouseY += " (minY)"
-            }
-          })
-          document.getElementById('preview').contentWindow.addEventListener("error", function(event) {
-            var error = event.error
-            app.seenError = false
-            if (error.type == "ImageLoadError") {
-              app.error = {message: "There is a problem with your image URL: " + error.url, lineno: code.substring(0, code.indexOf(error.url)).split('\n').length}
-            } else {
+        document.getElementById('preview').contentWindow.addEventListener("mousemove", function(event) {
```
stevekrouse: Is there a reason that you removed the try/catch around this section of code?

create.html
```diff
@@ -1474,55 +1520,50 @@ <h4 class="modal-body text-danger" :style="{display: !getID() ? 'block' : 'none'
...
+        })
+        document.getElementById('preview').contentWindow.addEventListener("error", function(event) {
+          var error = event.error
+          if (error.type == "ImageLoadError") {
+            app.error = {message: "There is a problem with your image URL: " + error.url, lineno: code.substring(0, code.indexOf(error.url)).split('\n').length}
+          } else {
               app.error = event
             }
-
+          app.showError()
```
stevekrouse: seems like this line and the line below need to be indented, yes?

create.html
```diff
@@ -700,60 +725,73 @@ <h4 class="modal-body text-danger" :style="{display: !getID() ? 'block' : 'none'
       methods: {
         getID: getID,
         showError: function(){
```
stevekrouse: It also might be worth renaming showError to a better name because it doesn't actually show the error (the error is shown if it is defined). This is more "proccessError" or something, yes?

create.html
madduccino: They actually shouldn't be indented, but the ones below it should be outdented. I changed the formatting of the else statement to make this more clear.

create.html
madduccino: I removed try/catch because the catch was not functioning. The catch only works in the woof js library when the error is thrown.

create.html
madduccino: Sounds good, done

create.html
madduccino: Changed it to replaceMisspelledWord?

PR #502, opened 2017-10-12
Auto-save on switch to fullscreen

gmarraro:


stevekrouse:

https://cdn.rawgit.com/stevekrouse/WoofJS/be27c0e9404701910dd1a11415301f0c0dbe412e/index.html

stevekrouse:
Nice, thanks for getting started on this! There are a few different cases that you'll need to handle...

| User | Project | Outcome |
--------|---------| --------------|
Logged out | Never Saved | ??
Logged out | Saved anonymously or by someone else | ??
Logged in | Never Saved | ??
Logged in | Saved by this user | ??
Logged in | Saved anonymously or by another user | ??

First things first, fill out this table with what you want to happen. 

Then write the code for each case and verify that it works on each one.

Let me know if you'd like more help :)

stevekrouse:
This project is actually more complicated than one might think because you can only open up a new page *after* the project has been successfully saved, thus you'll somehow need to pass in a callback to the 'save()` method or react to the fact that the project now has a name.

PR #507, opened 2017-10-17
freeze search bar in documentation pane

gmarraro:


gmarraro:
I'm having trouble with the search bar div laying on top of the div underneath it. I'm trying to get it to sit on top of the other blocks.

stevekrouse:

https://cdn.rawgit.com/stevekrouse/WoofJS/eac911d12b7d1274f9185a1f6b9af1648796982b/index.html

madduccino:
You should be able to get what you want by adding a margin to the div that wraps around everything that is equal to the height of the search bar, then adding a negative margin of the same value to the search bar. 

Also, when you click on a block, the search field looks a little long on the right. Maybe play with the widths of the back button and the search field. (I'd use percentages)

stevekrouse:
Thanks for the advice @madduccino!

@gmarraro thanks for starting to work on this one! 

I feel like the ultimate solution here will be copying Scratch's design for the blocks menu, which will be a potentially more involved redesign https://github.com/stevekrouse/WoofJS/issues/413. Let me know if you need more clarification.

PR #510, opened 2017-10-22
fix bug in error highlighting

madduccino:
There is a bug in removing error highlight under certain circumstances. Here's an example:

Error created on line 4:

screenshot 2017-10-21 at 9 06 51 pm

New code added which creates another error on line 3:

screenshot 2017-10-21 at 9 07 53 pm

Error on line 3 is fixed which resolves the original error that is now on line 6. As a result, `errorLineNumber` stays at 3 and the highlight on line 6 is not removed.

screenshot 2017-10-21 at 9 25 29 pm

I resolved this by removing the highlight from ALL lines each time `tryRunningCode` is run.

### DIFF
--- a/create.html
+++ b/create.html
@@ -1523,8 +1523,7 @@
     function tryRunningCode() {
       app.code = editor.getValue()
       app.error = undefined
-      var count = editor.lineCount(), i;
-      for (i = 0; i < count; i++) {
+      for (var i = 0; i < editor.lineCount(); i++) {
         editor.removeLineClass(i, 'text', 'error-highlight');
       }
       try {
--- a/create.html
+++ b/create.html
@@ -1523,8 +1523,9 @@
     function tryRunningCode() {
       app.code = editor.getValue()
       app.error = undefined
-      if (app.errorLineNumber) {
-        editor.removeLineClass(app.errorLineNumber - 1, 'text', 'error-highlight');
+      var count = editor.lineCount(), i;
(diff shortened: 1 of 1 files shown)


stevekrouse:

https://cdn.rawgit.com/stevekrouse/WoofJS/331e4608e20f937550bd0c0c0ee155bb3256ba1f/index.html

create.html
```diff
@@ -1523,8 +1523,9 @@ <h4 class="modal-body text-danger" :style="{display: !getID() ? 'block' : 'none'
     function tryRunningCode() {
       app.code = editor.getValue()
       app.error = undefined
-      if (app.errorLineNumber) {
-        editor.removeLineClass(app.errorLineNumber - 1, 'text', 'error-highlight');
+      var count = editor.lineCount(), i;
```
stevekrouse: This line confuses me. I would delete it and just make the next line:

```javascript
for (var i = 0; i < editor.lineCount(); i++) {
```

stevekrouse:
Great catch and solution 👍

create.html
madduccino: The reason I did this is so that it wouldn't have to calculate the line count each time the loop is run. Trying to optimize? Not worth it?

stevekrouse: Oh interesting - yeah that's probably an over-optimization but you can do it if you want.

The main thing that confuses me is the `, i;` part. We normally just define `var i` within the for-loop

Issue #511, opened 2017-10-24
Change the content of the default conditional statements so that they do not trigger an error when pasted into the editor.


MMarki:
Currently, conditional statements in Woof contain the text "/* condition */" inside their parentheses. This triggers a syntax error when pasted into the editor because JavaScript reads this as an empty conditional. I've seen a lot of students get anxious, worried, or annoyed when they paste in the default code and get an error message. Usually, they think they've done something wrong and they don't understand what has happened.

One possible solution would be to have a conditional statement like "1===1" instead. This would not trigger an error, and could lead to a valuable discussion with a student about booleans. The downside is that this portion of the statement is no longer explicitly labeled as a conditional.

stevekrouse:
This is a great idea! Maybe we do a more helpful/relevant default condition:

```javascript
if (mouseDown) {
  /* then do stuff here */
}
```

We could also update the picture of this block to have the mouseDown scratch block in the if-block in the documentation.

I'm not sure if I like this idea, but we could do both if we wanted...

```javascript
if(mouseDown /* replace in between the parenthesis with our own condition */) {
  /* then do stuff here */
}
```

PR #521, opened 2018-03-08
remove auto error from if block

madduccino:
Here are my thoughts and proposed solution for #511:

Students are often confused by the comments inside blocks and usually ignore the messages. So, something like this..

`if(mouseDown /* replace in between the parenthesis with our own condition */) {
  /* then do stuff here */
}`

...would probably lead to lots of errors.

I was thinking a shorter version that actually explains what is happening could be:

`if(/* something is */ true) {
  /* then do stuff here */
}`

This will probably lead to errors, too, which is why I decided to try adding example buttons. Students definitely don't read, and when they see a lot of text, I think it just leads to confusion. I thought the buttons might be a way of organizing all of the content so that it's not overwhelming while hopefully providing further opportunity to understand what each piece of code is actually saying.


stevekrouse:

https://cdn.rawgit.com/stevekrouse/WoofJS/871c1edee061bec5f53508579bf773942bc10732/index.html

stevekrouse:
Agreed this is an issue! 

## Quick solution

This is a clever solution: `if(/* something is */ true) { /* then do stuff here */ }`. One hesitation with this particular approach is that it will encourage students to do the silly `if (mouseDown == true)` thing... but I guess that's not the worst outcome, and it's definitely better than causing an error like we currently do. You may want to simply make this change separately and leave this pull request for examples.

As for "repeat instantly 10 times", you can also do this one separately as well by adding it as a new block of code underneath repeat().

## Example buttons

I like the idea of example buttons, but I find the interface you came up with a little confusing. Some ideas to improve the interface:

1. Put the example button closer to the Scratch image, above the code, to make it clear which block it goes with. Maybe it could be on the same line as the image. Or it could be next to the Copy button.
2. Make the example button smaller
3. Make it less colorful. Maybe use a bootstrap button default?
4. Make the button text uniform, such as "ex" or "example" because it should be obvious which block it is an example for.
5. Make the example button change the code for the example, not add a new code block. (You could also make the code block image change as well to have the parameters you add in the code... this makes me think it could be fun to have a Scratch to Woof interface)
6. Another idea: you could have different examples as "tabs" on the code interface, but may be a bit complicated...

PR #526, opened 2018-05-05
remove error from if blocks

madduccino:
Addresses first part of PR #521 - adds true to conditional blocks so that an error isn't automatically created.


stevekrouse:
💥

PR #529, opened 2018-05-19
requests save on fullscreen click then opens full screen automatically

leesan11:
I was able to test most conditions on the table and it was fully functional. The only one that I don't understand from the table is the last row. How do users save anonymously when logged in, on Woofjs?
Also, I was only able to test using a localhost environment (so not sure how _firebase_ took in data). I am not sure how to set up other environments. But there shouldn't be many problems because I used existing methods.

User | Project | Outcome
-- | -- | --
Logged out | Never Saved | Works
Logged out | Saved anonymously or by someone else | Works
Logged in | Never Saved | Works
Logged in | Saved by this user | Works (does not request save)
Logged in | Saved anonymously or by another user | I don't understand this part.

Cons of my code edits: 
- slightly tight coupling of functions save(), popout(), and makeclean() 
- created new key and boolean value requestfullscreen
-I just realized there will be a bug, if a student decides not to save and close the save-modal on fullscreen click. 
I'm thinking of creating a function on close modal click(https://github.com/stevekrouse/WoofJS/blob/master/create.html#L380). Do you mind if I do that? 

Please let me know, if I should make more changes. I apologize for making a new branch.

Thank you!


stevekrouse:

https://cdn.rawgit.com/leesan11/WoofJS/e33847e6f32c6551769ac15f6ac91841f657eda7/index.html

package-lock.json
```diff
@@ -1618,6 +1610,14 @@
             "strip-ansi": "3.0.1"
           }
         },
+        "string_decoder": {
```
stevekrouse: It doesn't look like this edit does anything. I imagine your build system made this change for some reason. 

It's not a big deal, but if you can revert this change to make the diff cleaner, that'd be great

create.html
```diff
@@ -710,6 +710,7 @@ <h4 class="modal-body text-danger" :style="{display: !getID() ? 'block' : 'none'
         migrateUsername: '',
         migrateChecked: true,
         migrateButtonState: "DISABLED", // NORMAL or DISABLED or LOADING
+        requestFullScreen:false
```
stevekrouse: I'd probably re-name this to `popoutAfterSave`

create.html
```diff
@@ -1264,6 +1265,10 @@ <h4 class="modal-body text-danger" :style="{display: !getID() ? 'block' : 'none'
     var makeClean = function() {
       app.status = "SAVED"
       currentHash = window.location.hash
+      //request fullscreen after save if requestfullscreen variable is true
```
stevekrouse: I'm definitely a fan of a comment here, but this comment merely describes what the code is already saying itself. It's too literal. It'd maybe putt the comment within the if-statement, above `popout()` and say:

```
// This save was initiated by the user clicked the popout button
// Now that the project is saved, popout the project in a new window
```

create.html
```diff
@@ -1264,6 +1265,10 @@ <h4 class="modal-body text-danger" :style="{display: !getID() ? 'block' : 'none'
...
+      if(app.requestFullScreen){
+        popout();
```
stevekrouse: This is probably the place to set `app.requestFullScreen = false`

create.html
```diff
@@ -1272,9 +1277,13 @@ <h4 class="modal-body text-danger" :style="{display: !getID() ? 'block' : 'none'
 
     var popout = function() {
       if (app.status == "SAVED"){
+        //change fullscreen request to false
+        app.requestFullScreen=false;
```
stevekrouse: As said above, you probably want this above and not here

create.html
```diff
@@ -1264,6 +1265,10 @@ <h4 class="modal-body text-danger" :style="{display: !getID() ? 'block' : 'none'
     var makeClean = function() {
```
stevekrouse: Maybe put this logic [here](https://github.com/stevekrouse/WoofJS/pull/529/files#diff-b6526a06bd2c9aea58ea7ea1f53f18f8R1375) instead and leave `makeClean` alone?

create.html
stevekrouse: Ditto for your other comments. Try making them higher-level

create.html
```diff
@@ -377,7 +377,7 @@ <h4 class="modal-title" id="saveModalLabel">Choose a name for your project</h4>
             <span id="illegalCharacter" class="text-danger" :class="{animated: projectNameLastKeyIllegal, shake: projectNameLastKeyIllegal}" style="display: inline-block" :style="{opacity: projectNameLastKeyIllegal ? '1' : '0'}">You cannot include {{ projectNameLastKeyIllegal }} in project names.</span>
           </div>
           <div class="modal-footer">
-            <button type="button" class="btn btn-default" data-dismiss="modal">Close</button>
+            <button type="button" class="btn btn-default" data-dismiss="modal" onclick="app.popoutAfterSave=false">Close</button>
```
stevekrouse: The idiomatic way to do this in VueJS is `@click="popoutAfterSave=false"`

create.html
```diff
@@ -1274,7 +1275,9 @@ <h4 class="modal-body text-danger" :style="{display: !getID() ? 'block' : 'none'
       if (app.status == "SAVED"){
         window.open("./full.html#" + getIDWithRevision(), "_blank")
       } else {
-        alert("You must save your project to open it in a new window.")
+        //when project is not in saved state, must set popoutAfterSave=true to run conditional popout() nested within save()>saveProjectFirebase
```
stevekrouse: One small stylistic comment: I prefer spaces after `//` (so before the "when" in this case)

This comment is still too tied to the code:

* "when project is not in saved state" -> this is already clear because it's in the else clause of `app.status == "SAVED"`
* "must set popoutAfterSave=true" -> this is literally the code below duplicated in a comment
* "to run conditional popout() nested within save()>saveProjectFirebase" -> this very literally describes where the code is below, and even that it is a conditional

An example of a higher-level comment: `// remember this save was initiated by clicking popout, so we remember to open the popout after saving is completed`  (to be honest, I don't love this comment either, but it's a in the right direction)

create.html
```diff
@@ -1274,7 +1275,9 @@ <h4 class="modal-body text-danger" :style="{display: !getID() ? 'block' : 'none'
...
+        app.popoutAfterSave=true;
```
stevekrouse: Please add spaces around the equal sign: `app.popoutAfterSave = true;`

create.html
```diff
@@ -1363,6 +1366,13 @@ <h4 class="modal-body text-danger" :style="{display: !getID() ? 'block' : 'none'
         window.location.hash = name
         setUserUploadImageListener() // sync user-uploaded photos for this project name
         myCode = firebase.auth().currentUser
+        // This save was initiated by the user clicked the popout button
```
stevekrouse: *clicking (sorry if that's my typo)

create.html
```diff
@@ -1363,6 +1366,13 @@ <h4 class="modal-body text-danger" :style="{display: !getID() ? 'block' : 'none'
...
+        // Now that the project is saved, popout the project in a new window
+        if(app.popoutAfterSave){
+          //popoutAfterSave is set, to prevent popout in all circumstances project is saved
```
stevekrouse: I'm not sure if you need a comment here. It's pretty self-explainatory. A higher level comment would be: `// reset popoutAfterSave after popout`, but that's basically what the code is already doing. (You could switch the two lines below to reset the boolean after the popout to make it clearer, but it's unnecessary.)

create.html
```diff
@@ -1363,6 +1366,13 @@ <h4 class="modal-body text-danger" :style="{display: !getID() ? 'block' : 'none'
...
+          app.popoutAfterSave=false;
```
stevekrouse: Again, spaces around the equal sign please

leesan11:
@stevekrouse Thank you so much for the feedback again! I believe that formatting and the style of the code is very important! I am starting to understand higher level comments a bit better, thanks to your examples!

For the latest edits I have:

- [x] changed comments as necessary
- [x] added spaces before and after = signs on assignments
- [x] changed onclick on @click 

Please let me know, if I should make further changes!

create.html
```diff
@@ -377,7 +377,7 @@ <h4 class="modal-title" id="saveModalLabel">Choose a name for your project</h4>
             <span id="illegalCharacter" class="text-danger" :class="{animated: projectNameLastKeyIllegal, shake: projectNameLastKeyIllegal}" style="display: inline-block" :style="{opacity: projectNameLastKeyIllegal ? '1' : '0'}">You cannot include {{ projectNameLastKeyIllegal }} in project names.</span>
           </div>
           <div class="modal-footer">
-            <button type="button" class="btn btn-default" data-dismiss="modal">Close</button>
+            <button type="button" class="btn btn-default" data-dismiss="modal" @click="app.popoutAfterSave=false">Close</button>
```
stevekrouse: This doesn't work. In view @click's you drop the `app.`. It's just `@click="popoutAfterSave=false"`. Next time make sure to check that it works before pushing

leesan11:
@stevekrouse So sorry about that, I should have tested it first. I just tested all the conditions with the new code edit. Thank you so much for all your time and guidance! I learned so much!

stevekrouse:
Thanks @leesan11! Glad to hear you had a positive experience. Are you interested in working on another issue? If so, I can help you find one that's priority and also not too complicated for a second issue - let me know

leesan11:
Thank you so much @stevekrouse ! I would love to work on another issue!

Issue #536, opened 2018-06-08
This.render is not a function?


benhtmn:
Occasionally woofjs freezes, requiring a reload, and the error message reads "this.render is not a function", while the little emoji thing glitches out, showing tons of different emojis. I don't have any specific case in which this error occurs (I'll make sure to keep an eye out in the future), but I've seen it multiple times so I figured I would make it known.

stevekrouse:
Yeah, I've seen this too. It's not Woof's best moment. If anyone can create a reproducible example of this bug and paste it below, that'd be a great start towards debugging it

stevekrouse:
+1 Another person chatted us to complain about this

proposal1220:
Actually, this happens to me on Chrome continually. I don't have to refresh the page. Very often I just increase or decrease the number of lines of codes by 1 or 2 and it fixes it. It seems to not be happening on my student's surface running chrome but it does on my surface running chrome.

proposal1220:
setBackdropURL("./docs/images/tomato-attack-backdrop.jpg")
setBackdropStyle("cover")

var unicorn = new Image({
  url: "./docs/images/unicorn.png",
  width: 120, 
  height: 80,
  
})
var bb8 = new Image({
  url: "./docs/images/BB8.png",
  width: 150, 
  height: 220,
})

That code with line 15 blank has " There's a problem.
this.render is not a function" the warning smileys cycle through them really fast, as if that error is happening repeatedly.

handatoe:
i have mostly gotten this.render is not a function when my code is really long. mostly i just delete a character and it will go. but as soon as i do something else it comes right back.

![screenshot 2019-03-01 at 5 28 31 pm](https://user-images.githubusercontent.com/44340761/53670530-0bd07200-3c49-11e9-96d2-f91ef0c288e1.png)

madduccino:
This is happening a lot now. Like in this project: https://woofjs.com/team.html#black-dragonfly

Not sure what's going on?

stevekrouse:
I never looked into this issue. Every time it seemed worth looking into (it was happening a lot), it would then seem to stop happening as much. It may be time to (finally) actually look into it...

As you know, the first essential step to solving any bug is to first come up with a reliable way to reproduce it. That is the problem with this issue. For example, the above link does not trigger the bug for me. I messed around with that code a bit: no issue.

[...]

PR #551, opened 2019-04-25
render error hotfix

madduccino:
@truthgoddess and I worked on this together to make the following adjustments that we think will remove most instances of a this.render error as described in issue #536 :

1. We changed the runCode call in the refresh button to debouncedRunCode. This takes away the ability for a user to run the code over and over again, reducing call stack errors to basically none.

2. Inside the runCode function, we changed the timeout from 10 to 100 to slow down the number of times tryRunningCode() is called.

3. We removed an extraneous (seemingly) runCode call here: https://github.com/stevekrouse/WoofJS/blob/master/create.html#L1510

In the future, we hope to a) find more efficient ways to run Woof in Woof (runs very well in JSBin for instance) and b) create more friendly error messages if a this.render error does pop up but also for every error.

Ways for you to see if the error is fixed:

Try reproducing the error in two ways on the current version and compare:

1. Press the reload button very quickly over and over again
2. Run your computer on a slower network (a teacher used a 3G connection or something on his Mac and was able to reproduce it)

### DIFF
--- a/create.html
+++ b/create.html
@@ -1582,9 +1582,12 @@
     }
 
     function runCode() {
-        var iframe = document.getElementById('preview');
+        
+        document.getElementById("preview").remove();
+        var iframe = document.createElement("iframe");
+        iframe.id = "preview";
+        document.getElementById("output").appendChild(iframe);
 
-        iframe.src = ""; // first, we clear the preview iframe
         setTimeout(function() {
 
             // add doctype to iframe


stevekrouse:

https://cdn.rawgit.com/stevekrouse/WoofJS/fdb30f1567c76de13aade679d8af82a232c227e6/index.html

create.html
```diff
@@ -1507,7 +1507,7 @@ <h4 class="modal-body text-danger" :style="{display: !getID() ? 'block' : 'none'
         setTimeout(function() {WoofJSLint()}, 1000); 
       }
     });
-    runCode()
+   // runCode()
```
stevekrouse: I think you are correct that we can remove this because the first evaluation of code is triggered when we load it from Firebase and put it in the editor, which triggers `debouncedRunCode`.

Just delete lines of code like this instead of commenting them out. (If we want them back, that's what git & version control are for.)

truthgoddess: Thanks. Will do that in the future!

truthgoddess:
Thanks for your thoughts and insights Steve. I've been trying to understand what has been called too many times, or why call stack errors were popping up and where. With Maddy's permission, I'll definitely work with her on integrating your ideas above. Thanks for helping us!

madduccino:
@truthgoddess, I'm happy to continue working on this with you! Thanks for all of your contributions here.

@stevekrouse in response to your feedback:

1. I don't know if kids are actually doing this, it just felt important to remove all possibilities of the error occurring.

2. I'm not sure "slow down the number of times" even makes sense. I think what I was trying to say is that it's waiting 10ms to add the woof.js script and tryRunningCode() is not going to run until that script loads. So, tryRunningCode() is itself delayed 10ms and thereby not being called instantly. It seems to me that tryRunningCode is [where the user's code is actually being added](https://github.com/stevekrouse/WoofJS/blob/master/create.html#L1576). 

If things are happening asynchronous, this would prevent a bunch of tryRunningCodes, and therefore a bunch of user code executions, from being called right on top of eacher other which might be triggering the error. It seems like tryRunningCode is also where a lot of error processing is happening.

I assume a lot of what I just wrote is incorrect. Any clarity you could provide here would be very much appreciated as I try to understand this!

3. I didn't make clear that removing the extraneous runCode() prevented an immediate error from occurring on page (or maybe iframe) load. Sorry for leaving it commented out, I thought I had ended up deleting it. I'm going to go ahead and make that change now. Would it be possible to merge these changes now and for us to follow up with more adjustments? There are students in our Monday class who haven't been able to use Woof for a couple of weeks (except in JSBin). I tested our local version out with a couple of students on Friday and no errors occurred during the entire class. I had another student act as a control and work on the live version where she continued to experience errors throughout the class.

stevekrouse:
I have no qualms with merging the change from `runCode` to `debouncedRunCode` and removing the first `runCode` call.

I am less comfortable with the 10ms to 100ms change. Ultimately, I think we need to refactor this section (and others like it) to an async/await/Promise style to get us out of "callback hell", so we can all understand what's going on...

But until then, let me explain what I think is happening:

`debouncedRunCode` waits until 1 second has past since it has last been called and then runs `runCode`. 

`runCode` first clears the contents of the iframe. It then sets a timeout for 10ms to add contents to the iframe. Here's where I think the issue is: [JavaScript's setTimeout is known to be unreliable and when the CPU is busy, timers drift closer to each other.](https://stackoverflow.com/a/29972322) 

Here's a possible timeline:

t=1000ms
runCode is called
setTimeout is set

t=2000ms
runCode is called
setTimeout is set

t=2500
the first setTimeout is evaluated

t=2501
the second setTimeout is evaluated

(The fact that we're also setting `script.onload = function() {` in the timeout probably is also relevant, but I won't go into that level of detail...)

As you can see, JavaScript's notion of a timeout is that something must not happen before the timeout interval but can happen any time afterwards. So maybe switching from 10ms to 100ms makes it less likely that calls will run into each other, but I imagine the bug will still happen. (It will also make the loading time of previews slower.)

One way to fix the overlapping timeouts issue is to clear old timeouts (via their timeout ID) before setting new ones, such as I prototyped with this code:

```
       // ensures that we clear old timeouts before setting up a new ones
        window.timeoutIds = window.timeoutIds || []
        window.timeoutIds.forEach(clearTimeout)
        
        iframe.src = ""; // first, we clear the preview iframe
        window.timeoutIds.push(setTimeout(function() {
```

Another thing that might help as I said above is 100% deleting (instead of merely clearing) the old iframe and replacing it with a new one every time. Mutating shared objects in asynchronous code is always a bad idea. Creating new independent copies usually is the way to go.

---

Speaking just about today, I can understand wanting to merge this fix because it seems to help. If that seems like the way to go, I won't object. Maybe just add a comment linking to this conversation and note that this code is a temporary fix that we don't 100% understand.

I would probably prefer if you instead made a tinyurl (or some other url shortener) to the rawgit link so kids could use the version you prefer without having to merge it into the code.

Speaking medium term, I think clearing old timeouts and using fresh iframes is the way to go.

Speaking longer term, I think a fuller refactoring may be in order for this section...

stevekrouse:
One other thing to add: as Woof is currently set up, you'll have to manually add these same changes to `/team` if you want to fix the issues there as well. 

The longer term solution is to make it so that /create and /team share more code so we don't need to copy and paste from one to the other...

madduccino:
I think you're right about deleting the iframe vs. clearing. Should've tried that on Saturday! 

I changed the delay back to 10ms, removed the iframe clear and replaced it with this:

```
       document.getElementById("preview").remove();
        var iframe = document.createElement("iframe");
        iframe.id = "preview";
        document.getElementById("output").appendChild(iframe);
```

No error thus far! 

I then changed the refresh button back to calling runCode and was able to mash on it without a problem.

Before I made any of these changes, I found this time that removing the "extraneous" runCode call made the backdrop not render initially. Maybe that's why it was added. I think we can keep it as it hasn't created an issue for me so far in testing with the iframe deletion happening.

Let me know if you'd like me to commit those changes!

madduccino:
I'm actually going to commit the above changes and have students work on this version rather than merging as you suggested.

madduccino:
(Also happy to work on escaping callback hell in the future)

truthgoddess:
Tested @madduccino changes on my gaming PC and on my Surface 4. The CPU overuse error is also corrected with this change. I could not get the render error once, even with overloading forever loops with 4 or 5 images infinitely created. 

Don't know if it is my place to say it, but great teamwork. ;) I hope @stevekrouse you see the same results and we can integrate this. Our Woof students will be soooo much happier for it.

stevekrouse:
Great, I merged it in. Seems like a good fix for now.

In the near term, clearing old timeouts is a good idea too, but ultimately a larger refactor is the way to go for this section.

(By the way, @madduccino I think you have permission to merge in pull requests or push to master yourself. If I'm not around to approve something and you have conviction about a change, you can always do it.)

madduccino:
Thanks for merging! And good to know. I really value your feedback and definitely prefer your approval, but I'll keep this in mind especially if something urgent comes up.

PR #552, opened 2019-05-02
nicer Woof JS errors

truthgoddess:
This update does a number of things that will help students who use WoofJS stay positive.

1) It removes the amount of red on the screen (chose light yellow which is easy to see, and is neutral with emotions) when displaying errors to decrease negative emotions when students are processing an error. 

2) Makes text for error messages black which contrasts with the light yellow better to help with reduced color view-ability, and to help make it easier for the user's eye to find the information it needs. (it also makes the other red marks easier to see)

3) Changes emojis for error messages to represent different cultures and genders, and changes the nature of those emojis to happy and content images to help students both feel they are represented, and help keep students calm when trying to troubleshoot.

4) Emojis are chosen from Twemoji (https://twemoji.twitter.com/content/twemoji-twitter/en.html) so that images we use are not copy-written (as the Apple graphics we were using were). Easy to see emoji sheet here: https://jollygoodcode.github.io/twemoji/

5) File structure was changed slightly. Made a new folder to store emoji images since the more there are, the more fun student's have with error messages (they may even create errors intentionally to see the different possibilities) and to keep our image file structure clean, and to allow original file names to stay intact for ease of replacement and identification with the Twemoji files themselves. 

In the future I may attempt to add text that encourages students, but this seemed to be a useful bunch of front end changes that can help students stay calm and feel happy when having to deal with the new complexities of text coding.


stevekrouse:

https://cdn.rawgit.com/truthgoddess/WoofJS/420483e0c9a907ce9c1f8e7245eedf601efccaa3/index.html

truthgoddess:
Forgot to mention, each svg are only around 2kb, which will save on memory and allow us to have as many emojis stored in our images file without creating bulk. Future additions to that emoji pool might be useful if students notice that images with error messages are interesting and varied. That slight gamification of errors might help students feel good about debugging at a time when many have to deal with that for the first time.

truthgoddess:
Thanks!

The reason behind the emoji's being random is because in the current version they are. In the current version, all the faces represent negative emotions. Student's don't process those often, but have a subconscious effect on them.

There are two subconscious effects this change is meant to help with:

1) Yellow face emojis represent to most of us white men. While there is nothing entirely wrong with that, one of Coding Space's initiatives is to make STEM and tech education more inclusive. The batch of emoji's I chose represent all the skin tones and genders available for a broad range of options. Students will not process these in a conscious way. But students of color and non-male students will see polite faces and emotions from characters that represent them in a field in which they may not often or don't. The inclusion of male and white skin tones is to keep including white boys as they always have been. Lastly, students who see representations of gender and skin types that are not their own will become more aware of the diversity around them and in this world. While the bodybuilding might appear to you to be the most representative and helpful, it is not for me. I chose it so that boys who like sports will feel represented. I chose other emojis so other people feel and have favorites and feel represented as well. 

2) Negative emotions expressed by emojis during errors are changed to positive and neutral ones. The point of that is to make the error process something less frustrating for students. A neutral or happy face next to an error message represents a friendlier character. I believe most students will see these characters as computer teachers telling them the error message. From what I've heard, the error descriptions up top were put there to make the error process for students more manageable. 

I do not know why the package-lock.json file was changed. I didn't do it. A quick search led me to think that the npm install instruction did that? 

I'm not entirely familiar with things, but how would I undo that? 

Thanks for considering this.

stevekrouse:
I think the safest and easiest solution is to remove the emojis entirely. They feel too random now to me, and I worry that they will confuse or distract people. All in all, I think this is a win for the product. The emojis were a bit of a silly gimmicky anyways, and it's especially good to remove them if they might be affecting students' emotions.

Let me get back to you on the package-lock...

stevekrouse:
It seems like the package-lock should be fine as is. My current understanding is that it keeps the team in sync regarding minor updates to our packages. It helps with debugging different behavior of our dependencies over time.

truthgoddess:
I hear you. No emojis might be better than what is there. How do you feel about the background color change of the top error? I think a neutral color there will reduce nervousness when seeing errors for students. It also seems to makes what is red stand out. Students seem to currently not read the left column errors at all as it is, but with more contrast that may utilize both. 

Thoughts? 

I can redo this branch with the no emojis and a background color that you think might work and resubmit it if you like. 

Perhaps a better way of inclusion of characters and representation might better be done on a Coding Space branch of WoofJS with better thought out character choices that guide the students in a friendly way.

stevekrouse:
The background color change is wonderful.

No need to redo the branch. On this branch you can simply delete the emojis and code that renders them, commit it, and push it. That will update this pull request. 

If you are worried about deleting forever your work on the emojis, you may be happy to hear that as long as you don't delete this branch, you can always [return to this work at this commit.](https://github.com/stevekrouse/WoofJS/pull/552/commits/420483e0c9a907ce9c1f8e7245eedf601efccaa3)

I think we'll want to keep the main branch of Woof for TCS, so let's talk on Friday about our process for discussing and planning product changes.

truthgoddess:
@stevekrouse Believe I changed it to what we agreed to. I copied the package.lock file from your file because I am not sure why it was different. I can't figure out what the difference is but have updated the relevant things to make this mergable I believe. Let me know if you see issues.

truthgoddess:
@stevekrouse @madduccino Since the original error box changes weren't added to /team.html, I'm going to put as a quick fix assigned to me so that the two are the same and once that is done we can make sure to update create and team at the same time until the include thing is figured out?

stevekrouse:
Great work! As far as the package-lock file, I am quite confused about it as well... I think what you did should be fine but we'll have to keep an eye on it going forward.

Good point as far as /team goes. It seems like we've been neglecting its updates for a while now. This is probably fine. I think it makes sense to **not** try and keep them in sync for now, because I should be addressing the root issue here via #555 soon

truthgoddess:
Thanks Steve! I agree with you re: /team. I personally think we should wait to update team until your part of the project that either uses and include to update one vs the other works. Team functionality doesn't seem to be utilized in many of my classes, and when it is, the kids are more distracted than anything, so my guess is that we can safely ignore the bugs there until our curriculum pushes younger students toward guided team coding. Besides, team coding can be done with one screen effectively, and if used, will probably be older kids on web. 

So, I'm going to assume that since both you and I agree, that unless I hear from Maddy, I will table the team updates, except when it is easy to update both at once?

stevekrouse:
👍

truthgoddess:
@madduccino Thoughts? It seems the /team issue is better done in the project health things that Steve does? As I mentioned to you yesterday, the fix to update /team doesn't seem as simple as we might have thought.

PR #557, opened 2019-05-20
clarify touching text

### LINKED ISSUE #544: Circle - Line collision detection
Right now the docs recommend using distanceTo(sprite) for circles, but if you use distanceTo for collisions with a line it does not work as intended. The collision is only detected if the circle comes the specified distance from the (x,y) point, but the Line is comprised of (x,y) and (x1,y1). 

Here is some code the recreates what I'm referring to:
```
var lineSprite1 = new Line({color: "pink", 
  width: 10, 
  x: -100, 
  y: 100, 
  x1: 10, 
  y1: 20
})

var circleSprite = new Circle({
  radius: 10, 
  color: "#444444", 
  x: 0, 
  y: 0
})

forever(() => {
  if (keysDown.includes('UP')) {
    circleSprite.y++;
  } else if (keysDown.includes('DOWN')) {
    circleSprite.y--;
  } else if (keysDown.includes('LEFT')) {
    circleSprite.x--;
  } else if (keysDown.includes('RIGHT')) {
    circleSprite.x++;
  }
  
  if (circleSprite.distanceTo(lineSprite1) < 10) {
    circleSprite.color = 'red'
  } else {
    circleSprite.color = '#444444'
  }
})
```

While using touching() does work, as the docs say it is not as precise at the distanceTo(sprite) function should be.

truthgoddess:
@madduccino and @stevekrouse 

The small change can allow us to close #544 . 

Not sure why the package-lock.json is changing again. Preliminary research tells me that this is something that is automatic. I think we'll be able to see if this is the same things for all of us and not just me if someone else creates any pull request. My guess is that someone should confirm this is ok, and then we should merge that change into the master so that we don't see this on every pull.

Also, for now, I'm not going to merge anything without approval. When I feel much more confident, I'll let either of you do the final approval. 

~Kate

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -1528,7 +1528,7 @@
                         tags: "touching intersecting overlap sensing"
                     }, 
                      {
-                        html: "<b>For circular sprites, <code>distanceTo()</code> will be more accurate than <code>touching()</code>.",
+                        html: "<b>For some sprites, <code>distanceTo()</code> will be more accurate than <code>touching()</code>.",
                         url: "./images/68747470733a2f2f692e696d6775722e636f6d2f733236773670632e706e67",
                         code: "sprite1.distanceTo(sprite2) < 100",
                         tags: "touching sensing distance"
--- a/package-lock.json
+++ b/package-lock.json
@@ -23,8 +23,8 @@
       "dev": true,
       "optional": true,
       "requires": {
-        "micromatch": "2.3.11",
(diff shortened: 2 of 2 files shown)


stevekrouse:

https://cdn.rawgit.com/stevekrouse/WoofJS/2ad788e97dba60d57acbad22f45f9670cb0f4368/index.html

PR #564, opened 2019-05-31
add tools menu items, and fix a few shortcuts, first try

truthgoddess:
Should close #448 if merged.

@stevekrouse I suspect you might have questions and comments on some things. Happy to go back and apply your suggestions.

PR #566, opened 2019-06-05
Non clickable tools tips

### LINKED ISSUE #563: Tips for Students commented at beginning of code.
See: https://woofjs.com/create.html#starting-text-test

While working on #448, ran into an issue in which some very cool tricks and shortcuts weren't easily or well placed as buttons in the tools menu. 

A solution is that we can put information as a comment in the code editor window as shown above. 

I will include reasons that CTRL-D is better placed here in a comment for #448, but I think for this issue, we can think about if this is good, what can be included, and then work on making it happen.

Pros to this:

Students will see comments in the code, and start wondering about it, why it exists, what to do with it, and what a comment is.

Students will eventually read and try out the shortcuts and commands themselves. Some of them are quite fun, and it will engage student's curiosity.

Links provided at the bottom encourage students to use resources that they need to understand Javascript. They can also reinforce that WoofJS is javascript. Currently, Piskel seems a very easy to work with and popular choice in my classes. The students who I point to that find it easy to use and makes Woof even more fun. 

@madduccino @stevekrouse What do you both think?

truthgoddess:
@madduccino @stevekrouse 

Put non-clickable tools, tips, and tricks in a menu which I have called Coding Tricks (but which is a placeholder in case we all have better ideas.) 

I moved the comments from the more blocks menu into this menu, since the more blocks menu is more focused on functions, and I believe we should in the future provide better tutorials on coding topics like functions, so the focused more blocks is a great place for that if we want to fill it out.

Added leftover commands that we wanted from #564. 

Tried to choose a color type that wasn't represented yet in Woof since this menu doesn't have a corresponding menu in Scratch. I'd arguing for a version of pink to add a color that attracts pink lovers, since the pink isn't represented anywhere, is clear, and can help brighten the experience for users. It is also a cultural color sometimes used in a disparaging way, and by associating it with WoofJS we help bring the color into a space that can be appreciated by everyone. 

That being said, I think the most important thing is choosing a color that isn't represented yet, and in the long term, I'm hoping our efforts make Woof more fun and immediately exciting, and I think a little splash of pink can be very exciting. 

Closes #563.


truthgoddess:
@nicolekelner Love your thoughts here too if you want.

stevekrouse:

https://cdn.rawgit.com/stevekrouse/WoofJS/51f384e08c86d1a7265b535fb33133c21948b950/index.html

truthgoddess:
I notice that I wrote backslash when I should have written forward slash, so I'll fix that before a merge is done, but I anticipate feedback, so I'll also fix it after the feedback.

truthgoddess:
Also notice that the horizontal lines are the color of the block in my branch. I think that would be better for all of our docs, rather than the grey. So, if anyone agrees or disagrees with that, I'm happy to adjust as necessary. Again, my thought is, WoofJS = more fun; students = excited about WoofJS

truthgoddess:
@stevekrouse @madduccino Bump! :)

Since camp is starting this week, and I feel that we are pretty organized, I'm probably going to have time to get back to curriculum and Woof. I'd love to your thoughts on this. Color wrong? Placement wrong? No color on bars? Not a good place for it? Anything at all helps. I don't like leaving things open. I'd love to move this forward step by step. 

Cheers.

madduccino:
I think it's great to have this information available! I'm not sure if it should be its own block, though. I'd vote for having a Coding Tricks option in the Tools menu and have a modal pop up when you click it.

truthgoddess:
Cool, I'll try that and update this!

stevekrouse:
Agreed with @madduccino. I might call it "More Tools". I worry that "Coding Tricks" might feel a bit too cute or childish.

PR #583, opened 2020-05-21
Update indexed repeat documentation

hallpell:


hallpell:
Have other stuff in this vein I want to add to documentation, holding off for a bit.

PR #590, opened 2021-04-12
Moves image uploads into Vue

danielpowell4:
This reworks the user-specific image uploads inside of 'images' inside
of 'sprites and backgrounds' to ensure all images (not just the first)
are shown to the user as options

Rather than living partially as 'raw' html via our block component, I
hoisted this up into the main app. This seemed to be a bit clearer in
terms of what `this.$refs` returned

It also provides a simple uploading indicator via disabling the button
and swapping out the text

I'll be honest that I am VERY NEW to Vue so if there is a better
way to accomplish these things, please LMK!


danielpowell4:
@madduccino pull this down and LMK if it works for you!

docs/index.html
```diff
@@ -2114,20 +2105,31 @@
...
-                            url: imageURL,
-                            code: 'var sprite1 = new Image({url: "' + imageURL + '"})',
-                            tags: "upload image picture",
-                            userUploadImage: true
-                        }
+                    // build user images, newest first
+                    images.reverse()
+                    var newImages = images.map(imageURL => ({
+                        url: imageURL,
+                        code: 'var sprite1 = new Image({url: "' + imageURL + '"})',
```
danielpowell4, line 2127: NOTE: all images are labeled as `sprite1` in this default code.

danielpowell4: Fixed in 424c1030c35 (next commit)

danielpowell4:
Per a conversation / review today, need to try harder on the sprite index labels

Something about the `reverse` isn't behaving as to be intended

create.html
```diff
@@ -1752,8 +1752,11 @@ <h4 class="modal-body text-danger" :style="{display: !getID() ? 'block' : 'none'
         .child(getID())
       ref.on('value', function(snapshot) {
         var addImagesFunc = () => {
-          var images = snapshot.val() || {}
-          scratch.contentWindow.app.setUserUploadImages(Object.values(images))
+          var imageValue = snapshot.val() || {}
+          var orderedKeys = Object.keys(imageValue).sort()
+          var images = orderedKeys.map(key => imageValue[key])
+
+          scratch.contentWindow.app.setUserUploadImages(images)
```
danielpowell4, line 1759: @madduccino LMK if this makes sense or not.

Was working for me (if I made a new project as nothing here back-populates the change to previous projects)

PR #591, opened 2021-04-13
Implement wait block

subtra3t:
I have implemented the `wait` block from Scratch (pun unintended). It's usage:

```js
wait(1, "seconds");
wait(1000, "milliseconds");
```

I have also changed docs/index.html to reflect the changes.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -1317,9 +1317,15 @@
                     Control: [{
                         url: "./images/68747470733a2f2f692e696d6775722e636f6d2f723677595838632e706e67",
                         code: 'after(4, "seconds", () => {\n  \/* do something here */\n})',
-                        description: 'There is no wait block in JavaScript. Instead, you can use after():',
+                        description: 'To execute a function after a certain amount of time, use after()',
                         tags: "wait after seconds control"
-                    }, {
+                    },
+			{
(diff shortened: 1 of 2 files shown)


tlangmo:
That implementation actually implements a busy wait in 
```
while (now < target) {
  now = Date.now();
}
```
That locks up the whole woofJS render loop and makes the UI stall. Recreating wait() for Scratch style is not trivial, but can be done with Javascript. 

Here is an example. Note that the IDE shows errors, most likely because the CodeMirror version is 7 years old. But the code works nonetheless
```
let rect = new Rectangle({color:"red"});

// a common pattern for a JS wait() is to wrap the setTimeout in a Promise
function wait(time) {
  return new Promise(resolve => setTimeout(resolve, time));
}
// we need an async function to allow await keyword
async function foreverWithWait(func) {
  await func();
  window.requestAnimationFrame(foreverWithWait(func));
}
// 
async function myAction() {
  rect.width = 100;
  // we need the 'await' keyword to pause execution
  await wait(1000);
  rect.width = 10;
  await wait(500);
}
// as close as we can get to Scratch serial syntax
foreverWithWait(myAction);
```

Note that **foreverWithWait** will be called outside the WoofJS render loop. In practice this should not be a big problem I assume.

PR #612, opened 2022-04-27
refactored the code variable

YTeyddie22:


YTeyddie22:
the string variable is in index.js

Issue #620, opened 2023-07-26
What if we just didn't use JSON syntax?


Wowfunhappy:
One of the biggest and consistent pain points for early WoofJS students is understanding the difference between JSON object notation and normal Javascript code. Children will confuse `:` with `=`, put normal code within JSON objects and vice versa, and so on.

What if the documentation just didn't contain any JSON?

In other words, instead of:

```
var unicorn = new Image({
  url: "./docs/images/unicorn.png",
  width: 120, 
  height: 80,
})
```

The documentation would have:

```
var unicorn = new Image();
unicorn.url = "./docs/images/unicorn.png"
unicorn.width = 120
unicorn.height = 80
```

Notably, the latter doesn't appear to actually work in WoofJS in the moment (which maybe should be tracked in a separate issue). The `width` and `height` take effect but the URL property does not, the image has the placeholder dog.

Regardless, I would much prefer if we used the latter in documentation, I think it would help a lot of students. We _would_ need to really encourage students to have line breaks between the "sprites" or the code could get very difficult to read.

hallpell:
This URL code doesn't currently work because you need to call setImageURL. Though making it so changing the url call that automatically is pretty easy and I was planning on doing that anyway when I had time.

I do prefer it the way you propose - creating the objects to pass to the constructor is only a positive if we're aiming to teach that distinction between object construction and modifying it after the fact, which is not something I feel the need to teach. A quick skim through the documentation seems like we could probably pull all of it out. The only parts that I need to think through more is what to do with the extra characteristics after creating the sprite. In the text part of the documentation, it has stuff like:
`x: 50, y: -100`
`fontFamily: 'arial'`
which in the new paradigm would be somewhat replicated in motion and looks. Which I think is fine, but mulling it over a bit more before making a full shift.

Wowfunhappy:
> A quick skim through the documentation seems like we could probably pull all of it out. The only parts that I need to think through more is what to do with the extra characteristics after creating the sprite. In the text part of the documentation, it has stuff like:
`x: 50, y: -100`
`fontFamily: 'arial'`

I may not be understanding, I don't see why you wouldn't write this as:

```
textSprite1.x = 50
textSprite1.y = -100
textSprite1.fontFamily = 'arial'`
```

If there's anything that ends up just being duplicative... personally, I've _long_ thought there's "too much" in the documentation and I'm all for cutting it down.

Issue #618, opened 2023-05-23
Sprites with NO ROTATE will still have their collider rotate


hallpell:

hallpell:
Additionally, if a sprite's collider isn't on the screen, we won't render it, which can cause sprites to "pop" in since they're not drawn even if they should be. Make this more pressing.

Issue #617, opened 2023-02-28
No Rotate on Line Sprites


hallpell:
We allow changing the rotation style of Line Sprites to No Rotate, which really hobbles their usefulness. We should probably not allow this and throw an error if you try. 

Based on observing the effects, I suspect we're implementing this with a Rectangle Sprite under the hood, and when you set No Rotate it makes x1 and y1 change the length of the Rectangle Sprite, but then we can't rotate it so it just hangs out.

Issue #609, opened 2022-01-18
Error message for improper URLs are attributed to incorrect line


hallpell:
In this example:

https://woofjs.com/create.html#hallpell_url-error

the error occurs in line 9 (where no URL is provided to setImageURL), but Woof reports the error on line 1. Need to process this error differently in some way.

hallpell:
It looks like we've already tried to deal with this by scanning through the code (regex-style) to look for Empty URL Types (https://github.com/stevekrouse/WoofJS/blob/42a2c231df3ddf922587df68eb4c413f11d343c8/create.html#L1711), but we're missing some cases. Three main cases that I think are missing are:
- I don't think it ignores whitespace
- If no argument is provided (which appears as 'undefined' once it gets inside woof.js and thrown as an ImageLoadError)
- If an empty variable is provided as an argument, we're probably not going to be able to track it down regex-style, and would need to track the error from where it was thrown, which requires upgrading our architecture a bit.

I should be able to clean up the first two cases a bit fairly easily, so should be able to get that done soon.

hallpell:
(Also setBackdropURL has an error message if a string isn't provided as the first argument, which should be added to setImageURL as well)

hallpell:
Further investigation of this makes me not trust how we handle errors in Woof. I believe when we try to GET an image that doesn't exist, we throw two errors. The 404 error seems to not be caught properly, and is making it hard for me to properly track down how the errors are propagating.

PR #621, opened 2023-09-01
Documentation Update + Minor Tweaks

### LINKED ISSUE #610: Polygon's length attribute is a misleading label
The length attribute of polygons makes me think it would be the side length of the polygon that we draw. But not so! It actually represents the distance from the center of the polygon (the X,Y coordinates of it) to the vertices of the polygon. (In the extreme case of sides=infinity, this would be the radius of the circle). I definitely don't want to change the name of length, and I don't think I want to change how length works (i.e. changing the size of polygons). I think I'm just going to add to the documentation explaining this to the polygon documentation and call it good, but will mull over for a bit before putting that in.

hallpell:
Main update is to change documentation to no longer use JSON objects for constructors, as per #620 . (I also added some context to Polygon lengths to close #610 )

Various tweaks:
Image Sprites and Sound Objects now have a .url property that behaves like I'd expect (having a get and set property, and setting will update the corresponding media).

Various rotation style stuff was cleaned up (#617 and #618).

My initial investigations and half-fix for #609 are also in this PR, but the errors from that are only marginally better than what we had before (previously it claimed the error was on line 1, now we don't give a line number).

### DIFF
--- a/create.html
+++ b/create.html
@@ -1767,10 +1767,13 @@
             iframe.contentWindow.document.body.appendChild(script);
 
             script.onload = function() {
-                // when the woof.js library loads, run the user's code
-                var evt = document.createEvent('Event');
-                evt.initEvent('load', false, false);
-                iframe.contentWindow.dispatchEvent(evt);
+		// when the woof.js library loads, run the user's code
+		// firefox automatically fires these events, so don't do it
+		if (!navigator.userAgent.match(/firefox/i)) {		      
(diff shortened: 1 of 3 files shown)


Wowfunhappy:
Yay, this great!

I guess since this is kept in a different repo it will move separately, but I wanted to flag that the example code in Cloning Tutorial needs to be updated to remove JSON objectives for constructors.

